PR #70, opened 2026-06-08
feat(pricing): build-time pricing data from models.dev (no runtime network)

### LINKED ISSUE #45: FYI: Leveraging `mmmodels` for better cost estimates?
Just a quick note to put this on your radar, in case it might come in handy as tool or reference implementation for estimating costs of different providers/models...  

> `mmmodels` is a CLI for browsing, filtering, and exploring AI models from hundreds of providers. Built for both humans and agents.

The author is the founder of both _Command Code_ and _LangBase_ so safe to assume his TypeScript code in this repo is pretty solid.
- Repo: https://github.com/ahmadawais/mmmodels
- Announcement: https://x.com/MrAhmadAwais/status/2031476826032509300

VasiHemanth:
Closes #45.

## What

The inline pricing table in `backend/pricing.py` covered ~50 models; models.dev covers 2000+. This unit broadens and freshens coverage **without** introducing any user-machine network call — TokenTelemetry stays local-first.

## How

- **`backend/pricing_sync.py`** — dev/CI-only script (never imported by the app). Fetches `https://models.dev/api.json` and regenerates `pricing_data.json` in our `{in, out, cached_read}` shape. Resilient: clear error if offline; refuses to overwrite with empty/garbage data; sanity-bounds rates.
- **`backend/pricing_data.json`** — committed snapshot generated by the script: **2328 models, 4840 provider-keyed entries**. Shipped with the package; users get fresh prices via the normal version update.
- **`backend/pricing.py`** — at import, reads the bundled JSON from `Path(__file__).parent` (never `~/.tokentelemetry`) and overlays it onto the inline `PRICING` / `PRICING_BY_PROVIDER` dicts. **Inline entries stay authoritative**; the file is also the fallback if absent/malformed. `calculate_cost()` and `_normalize_model_id()` are untouched (a sibling unit owns the rate math). **Zero network I/O at import or runtime.**
- **`.github/workflows/pricing-sync.yml`** — weekly job that runs the sync and opens a PR for maintainer review.
- **`backend/test_pricing_data.py`** — overlay applies; inline entries win; missing/garbage/wrong-shape file → silent fallback; import does no network I/O.

## Verification

- `python -m pytest backend/ -v` → 19 passed (9 new).
- Import with `socket.socket` blocked succeeds; `len(PRICING)` grows ~50 → 2350.
- `uvicorn main:app --port 8000` + `curl /analytics` → costs still compute.

## Note

`calculate_cost`'s fuzzy substring match now iterates the larger table; with short overlay ids (e.g. `o1`, `o3`) a spurious substring hit is theoretically possible, mitigated by its existing longest-key-first ordering. That function is owned by a sibling unit and intentionally left unchanged here.

### DIFF
--- a/.github/workflows/pricing-sync.yml
+++ b/.github/workflows/pricing-sync.yml
@@ -1,0 +1,48 @@
+name: Pricing sync (models.dev)
+
+# Refreshes backend/pricing_data.json from models.dev and opens a PR for review.
+# This is the ONLY place pricing data is fetched over the network — it runs in
+# CI, never on a user machine. TokenTelemetry stays local-first: users get fresh
+# prices through the normal version update, with no new outbound call.
+
+on:
+  schedule:
+    - cron: "0 6 * * 1"  # weekly, Mondays 06:00 UTC
+  workflow_dispatch: {}
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  sync:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.11"
+
+      - name: Regenerate pricing_data.json
+        run: python backend/pricing_sync.py
+
+      - name: Open PR with refreshed pricing
+        uses: peter-evans/create-pull-request@v6
+        with:
+          commit-message: "chore(pricing): refresh pricing_data.json from models.dev"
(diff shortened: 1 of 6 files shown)


VasiHemanth:
Superseded by #74 — consolidated into a single cost/token-accuracy PR (reviewed and validated together, including the new per-agent billing-mode work). This branch's change is included there as its own commit.

PR #73, opened 2026-06-08
feat: power & subscription cost config for local/subscription models (#49)

VasiHemanth:
## Unit 4 (#49) — power & subscription cost for local/subscription models

`backend/pricing.py`'s table maps model -> per-token USD, which is wrong for:
- **local models** (ollama / llama.cpp / vLLM) — no API bill; real cost is electricity;
- **flat subscriptions** (Ollama Cloud Pro, proxied gateways) — billed monthly, so per-call cost should be 0.

Today both fall through to a wrong rate or `_default`.

## Changes

- **new `backend/power_config.py`** — loads `~/.tokentelemetry/power.json` (`loadWatts`, `costPerKwh`, `subscriptionEndpoints`) with sane defaults; never raises on a missing/malformed file. Exposes `load/save_power_config`, `is_subscription_endpoint`, `electricity_cost`, and gating helpers.
- **`pricing.calculate_cost`** — new optional `endpoint` / `tok_per_sec` kwargs (fully backward compatible). Early branches in the LOOKUP region (above the rate-math owned by a sibling unit):
  - endpoint matches a subscription endpoint -> `0.0`;
  - unknown model **and** the user set an explicit power figure -> electricity cost `(output_tokens / tok_per_sec) * loadWatts / 3_600_000 * costPerKwh` (default 30 tok/s).
- **`backend/main.py`** — Hermes scan now selects `billing_base_url` (PRAGMA-guarded so legacy DBs don't break) and passes it as `endpoint`; new `GET/PUT /config/power` endpoints with validation in `power_config` (no raw errors surfaced).
- **`backend/test_power_config.py`** — defaults / malformed tolerance, subscription -> 0, local electricity pricing, and a regression test that subscription-only configs do **not** re-price unknown cloud models.
- **`UPDATE.json`** — new release entry (required by the pre-push hook).

## Notable design decision (from self-review)

Electricity pricing is gated on `local_power_enabled()` — an actual `loadWatts`/`costPerKwh` written to disk — **not** merely "a config file exists". Otherwise a user who configured only `subscriptionEndpoints` would have every *unknown cloud* model silently re-priced as near-zero electricity, under-reporting real API spend. Regression test included.

## UI scope

Scoped to backend + `GET/PUT /config/power` (per the unit's allowance to skip the UI given the Next.js breaking-changes caveat). A Settings panel can land in a follow-up.

## Verification

- `pytest backend/` — 20 passed (9 new in `test_power_config.py`).
- Ran uvicorn; `GET/PUT /config/power` round-trip works, bad values skipped gracefully, `/analytics` still loads Hermes sessions with the new column.

### DIFF
--- a/UPDATE.json
+++ b/UPDATE.json
@@ -1,6 +1,20 @@
 {
   "releases": [
     {
+      "tag": "2026-06-08",
+      "title": "Accurate cost for local & subscription models",
+      "highlights": [
+        {
+          "title": "Local models priced by electricity, not fake API rates",
+          "description": "Models you run yourself (Ollama, llama.cpp, vLLM) have no API bill — their real cost is power. Set your machine's watts under load and your kWh rate, and TokenTelemetry estimates the electricity cost instead of inventing a per-token charge."
+        },
+        {
+          "title": "Flat subscriptions now show $0 per call",
+          "description": "Endpoints you pay for monthly (Ollama Cloud Pro, a proxied gateway) are billed by subscription, not per token. List them as subscription endpoints and their sessions stop inflating your per-call cost totals."
+        }
+      ]
+    },
+    {
       "tag": "2026-06-07",
       "title": "Update check — now visible and yours to control",
       "highlights": [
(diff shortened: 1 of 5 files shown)


VasiHemanth:
Superseded by #74 — consolidated into a single cost/token-accuracy PR (reviewed and validated together, including the new per-agent billing-mode work). This branch's change is included there as its own commit.

PR #81, opened 2026-06-08
feat: opt-in remote/tailnet access via --host and --allowed-origins

JoshHobbs:
## Summary
Make the dashboard reachable from another device (a headless home server, a Tailscale tailnet) via opt-in, env/flag-driven config. Default behavior is unchanged: loopback-only.

## Type of change
- [x] New feature

## Motivation
TokenTelemetry assumes it runs *and* is viewed on `localhost`, so viewing it from another machine currently requires editing source in four places:
- the backend binds `127.0.0.1` (hardcoded);
- CORS allows only `localhost` / `127.0.0.1` origins;
- the frontend bakes an absolute `NEXT_PUBLIC_API_BASE`, so a remote browser ends up calling *its own* localhost;
- Next 15's `next dev` refuses to serve its JS chunks / HMR to a non-localhost origin, so the page loads as a blank shell.

## What changed
- **`bin/cli.js`**: new `--host` (`TT_HOST`) and `--allowed-origins` (`TT_ALLOWED_ORIGINS`) flags. The backend is launched with `--host`; the frontend is handed the API *port* (`NEXT_PUBLIC_API_PORT`) instead of a baked absolute base. A concrete (non-wildcard) `--host` is auto-folded into the allow-list, so `--host <ip>` works without also repeating the address in `--allowed-origins`.
- **`backend/main.py`**: the uvicorn bind resolves `--host → TT_HOST → 127.0.0.1` (mirrors the existing `_resolve_port` / `TT_API_PORT` / `--port` convention). The CORS allowlist is built from loopback + `TT_ALLOWED_ORIGINS`.
- **`frontend/src/lib/api.ts`**: `API_BASE` is derived from `window.location` + the API port at runtime, so a single build works on localhost, a LAN IP, or a tailnet host. An explicit `NEXT_PUBLIC_API_BASE` still overrides.
- **`frontend/next.config.ts`**: `allowedDevOrigins` is populated from `TT_ALLOWED_ORIGINS` (Next 15 dev-server requirement for non-localhost origins).
- `UPDATE.json` release entry + a one-line `.claude/CLAUDE.md` correction.

`TT_ALLOWED_ORIGINS` is the single source of truth feeding both the backend CORS allowlist and Next's dev origins. Remote exposure is strictly opt-in — with no flags the behavior is byte-for-byte the old loopback-only experience.

## How to test
1. **Default unchanged:** `./start.sh` → still serves on `http://localhost:3000` exactly as before.
2. **Remote:** on a server, `./start.sh --host 0.0.0.0 --allowed-origins <your-host>`, then open `http://<your-host>:3000` from another device — the dashboard loads and data populates.
3. **Security:** with no flags, a cross-origin request (`curl -H "Origin: http://evil:3000" localhost:8000/...`) is still rejected by CORS and the bind stays `127.0.0.1`.

## Checklist
- [x] I tested this locally with `./start.sh` (and end-to-end on a real Tailscale host)
- [x] CHANGELOG (`UPDATE.json`) updated

## Related issue
N/A


VasiHemanth:
Thanks for this, @JoshHobbs — the opt-in remote/tailnet access is exactly what we needed. I pushed two commits onto this branch (via maintainer edits) to harden it before merge:

**1. Access-token gate (`c278cce`)** — binding to a non-loopback `--host` previously exposed an *unauthenticated, writable* API to the network (CORS only restrains browsers, not direct `curl`/script clients). Now remote requests require a bearer token; **loopback stays exempt** and the default `./start.sh` is byte-for-byte unchanged. A token auto-generates on a remote bind (printed once), with `--insecure-no-auth` to opt out on a fully trusted tailnet. The gate is registered *before* CORS so CORS stays outermost (preflight + 401 still carry CORS headers).

**2. Scan-to-connect (`6aab12a`)** — a one-time `?token=` URL bootstrap + a "Connect a device" QR panel in Settings, so you can open the dashboard on a phone without typing the IP or token. The QR/link is served **loopback-only**, so the token can't be fetched from a remote device.

Tests: `python backend/test_remote_auth.py` (13 cases — gate on/off, loopback-exempt, header + `?token=`, preflight, loopback-only `/remote-access`); frontend typechecks clean.

Could you give it a spin on your setup — especially the Tailscale path you originally validated — and confirm it still works for you?
- **Default:** `./start.sh` → loopback only, no token (unchanged).
- **Remote:** `./start.sh --host 0.0.0.0 --allowed-origins <your-host>` → token printed once; open the connect link (or scan the QR in Settings) from another device.

Once you're happy, we'll merge. Thanks again! 🙏

PR #99, opened 2026-06-11
fix(summarizers): launch CLIs by resolved path and feed claude prompt via stdin

### LINKED ISSUE #98: Failed generate trace summary using Claude code
## Description
When attempting to generate a trace summary using Claude Code, the summarizer fails to launch with a WinError 2 exception. The application is unable to find the claude executable on the system.

## Error Message
```
Summarizer failed
failed to launch claude: [WinError 2] The system cannot find the file specified
```

## Steps to Reproduce
1. Configure Claude code as the trace summary backend
2. Make sure Claude code is available in the terminal:
	Image
3. Attempt to generate a Trace Summary using Claude Code.
4. Observe the error message in the Trace Summary UI.
	Image

## Environment
OS: Windows 11 Pro
TokenTelemetry Version: main branch
Claude Code CLI Version: 2.1.173

## Expected Behavior
The summarizer should successfully launch claude, process the trace, and generate a summary without throwing a system path error.

---
_Sent from TokenTelemetry · /sessions/aae91471-a90b-449b-a301-b86042ea265a_

fanyi-zhao:
## Summary

Fixes two Windows-only failures in the CLI summarizer launch path, surfaced when generating a trace summary with the Claude Code backend (#98).

### 1). `failed to launch claude: [WinError 2]`
On Windows the npm-installed coding CLIs are batch shims (`claude.cmd`, no `claude.exe`). `BaseSummarizer.is_available()` uses `shutil.which`, which honours `PATHEXT` and finds the shim — but `run_cli` launched a bare `subprocess.run([claude, …])`, which goes through Win32 `CreateProcess` and only resolves `.exe`. So the picker showed Claude as available while the actual launch raised `[WinError 2]`, even though `claude` works in the shell.

**Fix:** resolve `cmd[0]` to its full path via `shutil.which` before launching (`_resolve_executable` in `base.py`). On macOS/Linux `which` returns the plain executable path, so it's a no-op. Benefits all four CLI adapters (claude/codex/gemini/qwen), since they all route through `run_cli`.

### 2). `claude exited 1: The command line is too long.`
With 1) fixed, the next failure appeared: the claude adapter passed the prompt as a `-p <prompt>` argv, and trace briefs are routinely tens-to-hundreds of KB — well past cmd.exe's ~8 KB command-line limit (the `claude.cmd` shim runs under cmd.exe).

**Fix:** feed the prompt over **stdin** (`claude -p` with no positional), matching what the codex/gemini/qwen/antigravity adapters already do. The prompt no longer appears on the command line at all, so the limit can't be hit regardless of brief size.

## Testing
- Verified `run_cli([claude, --version])` now resolves to `…\claude.CMD` and returns `2.1.173` (previously `[WinError 2]`).
- Verified `claude -p` with the prompt on stdin returns the expected JSON `result` envelope.
- Confirmed the other CLI adapters were already stdin-based, so Claude was the lone outlier.
- Confirmed the trace summary now works:
	image

Fixes #98

### DIFF
--- a/backend/summarizers/base.py
+++ b/backend/summarizers/base.py
@@ -39,6 +39,25 @@ def _ensure_cwd() -> str:
     return str(SUMMARIZER_CWD)
 
 
+def _resolve_executable(name: str) -> str:
+    """Resolve a bare CLI name to its full path on PATH, honouring PATHEXT.
+
+    On Windows the coding CLIs are installed by npm as batch shims
+    (``claude.cmd``, ``codex.cmd``, …) — there is no ``.exe``. ``shutil.which``
+    finds them because it consults ``PATHEXT``, but a bare-name
+    ``subprocess.run(["claude", …])`` with ``shell=False`` goes straight to
+    Win32 ``CreateProcess``, which only appends ``.exe`` and so raises
+    ``FileNotFoundError`` / ``[WinError 2]`` even though the CLI works in the
+    shell. Handing subprocess the fully-resolved ``…\\claude.cmd`` path launches
+    correctly (CreateProcess runs the batch shim by full path). On macOS/Linux
+    ``which`` just returns the plain executable path, so this is a no-op there.
+
+    Falls back to the original name when nothing is found, so the existing
+    "failed to launch" error still surfaces with the same wording.
+    """
+    return shutil.which(name) or name
+
+
 def run_cli(
     cmd: list[str],
     *,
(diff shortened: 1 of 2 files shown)


VasiHemanth:
Hey @fanyi-zhao — thank you for this! Really appreciate you not just filing the bug but digging all the way to the root cause, identifying the second failure hiding behind the first, and shipping a clean fix for both in one go.

The `_resolve_executable` approach is exactly right — minimal, cross-platform safe, and it benefits all four CLI adapters at once. And routing the prompt through stdin is the correct long-term fix regardless of OS; it's consistent with how codex/gemini/qwen already work and removes the command-line length as a failure mode entirely.

I validated both fixes on macOS (full-path resolution, stdin on small + 50 KB prompts, error message preservation) — everything looks solid. Merging now.

PR #127, opened 2026-07-07
fix: compute cache_hit_pct from cumulative cache reads, not the per-session high-water mark

### LINKED ISSUE #126: cache_hit_pct understates Claude cache hit rate — mixes per-session high-water mark with cumulative input
## Summary

`/analytics` reports a far-too-low `cache_hit_pct` for Claude (and any scanner that stores `_cached_sum`), and the understatement grows with session length. On my corpus (421 sessions), Claude showed **~70%** when the true rate is **~99.9%**.

## Root cause

The Claude-style scanners intentionally store two different semantics (see the comment `# cached = unique cached-prefix size (high-water-mark), NOT per-turn sum` in `_scan_sessions_sync`):

- `tokens["input"]` — **cumulative** uncached input summed across every turn
- `tokens["cached"]` — **max** cache-read seen in any single turn (so `total`/cost reflect unique context footprint)
- `tokens["_cached_sum"]` — the actual cumulative cache reads

`_cache_hit_pct(input, cached)` in `/analytics` divides the HWM by the cumulative input — a per-session maximum over a per-turn sum. A 50-turn session that re-reads a ~200k prefix every turn counts ~200k in the numerator but every turn's uncached tokens in the denominator, so long agentic sessions drag the rate down hardest.

## Verification

Recomputed straight from Claude Code session JSONLs (sum of `cache_read_input_tokens` vs sum of `input_tokens`):

| Model | reads/(input+reads) | as displayed |
|---|---|---|
| claude-opus-4-8 | 100.0% | 75.1% |
| claude-sonnet-5 | 99.9% | 75.9% |

## Fix

PR to follow: aggregate a `cache_reads` field per agent (prefer `_cached_sum`, fall back to `cached` so scanners without it are byte-identical to today) and feed that into `_cache_hit_pct` for the per-agent and overall metrics. No changes to `cached`/`total`/cost semantics.

tom-swift-tech:
Fixes #126.

## What

`/analytics` fed the per-session **high-water-mark** `cached` field (unique cached-prefix size) into `_cache_hit_pct()` alongside **cumulative** input, so the reported hit rate shrank as sessions grew longer. On a real Claude Code corpus this displayed ~70% for models whose true hit rate is ~99.9% (verified by recomputing from the session JSONLs directly — see the issue for numbers).

## How

- `by_agent` aggregation gains a `cache_reads` field: prefers the scanner's cumulative `_cached_sum`, falls back to `cached` — so agents whose scanners don't track `_cached_sum` (codex, gemini, antigravity, …) produce byte-identical output to today.
- Per-agent and overall `cache_hit_pct` now compute from `cache_reads`.
- `cached`, `total`, and all cost math are untouched — the HWM semantics stay load-bearing for footprint totals and pricing.
- `_cache_hit_pct()` docstring now states the cumulative-reads contract.

## Verification

Before/after on my live data (421 sessions): claude 69–76% → **99.9%**, overall → **99.7%**; codex/gemini/antigravity unchanged (91.9 / 28.2 / 45.0). The claude figure matches an independent recompute summing `cache_read_input_tokens` / `input_tokens` straight from the JSONL transcripts.

### DIFF
--- a/backend/history_store.py
+++ b/backend/history_store.py
@@ -36,7 +36,7 @@ from tt_paths import data_dir
 
 _log = logging.getLogger("tokentelemetry.history")
 
-SCHEMA_VERSION = 1
+SCHEMA_VERSION = 2
 
 # Sub-dicts folded into ``ecosystem_json`` and expanded back out on read. These
 # are the keys the analytics aggregation + delegation views consume beyond the
@@ -82,6 +82,7 @@ def _migrate(con: sqlite3.Connection) -> None:
                 input           INTEGER DEFAULT 0,
                 output          INTEGER DEFAULT 0,
                 cached          INTEGER DEFAULT 0,
+                cache_reads     INTEGER DEFAULT 0,
                 total           INTEGER DEFAULT 0,
                 cost            REAL    DEFAULT 0.0,
                 tok_per_sec     REAL,
@@ -115,6 +116,20 @@ def _migrate(con: sqlite3.Connection) -> None:
             );
             """
         )
+        con.execute(f"PRAGMA user_version={SCHEMA_VERSION}")
+        con.commit()
+    if ver < 2:
+        # v2 adds `cache_reads`: the CUMULATIVE cache-read sum per session
+        # (Claude-style scanners' `_cached_sum`). Distinct from `cached`, which
+        # is the per-session high-water mark. The analytics cache-hit metric
+        # needs cumulative reads; without persisting it, sessions served from
+        # this store fall back to the HWM and understate the rate (issue #126).
+        # ALTER always appends the column; ``ver < 1`` fresh DBs already have it
+        # from CREATE TABLE, so tolerate the duplicate-column error.
+        try:
+            con.execute("ALTER TABLE sessions ADD COLUMN cache_reads INTEGER DEFAULT 0")
+        except sqlite3.OperationalError:
+            pass
         con.execute(f"PRAGMA user_version={SCHEMA_VERSION}")
         con.commit()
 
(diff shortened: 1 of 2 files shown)


VasiHemanth:
Thanks for catching this and writing it up so precisely, @tom-swift-tech — the root cause (HWM `cached` over cumulative `input`) and your `_cached_sum` switch are exactly right, and I verified the arithmetic against a raw recompute from the Claude JSONLs (summing `cache_read_input_tokens` / `input_tokens`): overall true ~99.9%, matching your fix.

I pushed one follow-up commit onto your branch (`8a45236`) to close a gap your change surfaced, so this is complete rather than partial:

**The gap.** Your fix reads `st.get("_cached_sum")`, but only the live file scan produces `_cached_sum`. `history_store` never persisted it — the schema has `cached` but no cumulative column, and `_rehydrate` didn't return one. So any session served from the durable store fell back to the HWM and still understated:
- **today-excluding date windows** — `get_analytics` answers those entirely from SQLite (per the comment at `main.py:6087`), so the buggy number came back;
- **pruned sessions** — the store exists precisely to outlive transcript pruning, and those rows are SQLite-only, so they'd stay understated even on the default view, worsening over time.

Your all-time verification passed because that day's full re-scan happened to override every stored row with a fresh `_cached_sum`.

**The follow-up commit.** Adds a `cache_reads` column (schema v2 + an `ALTER` migration for existing DBs), writes it from `_cached_sum` with a HWM fallback for scanners that don't track a sum, and rehydrates it into `tokens._cached_sum` — so your analytics aggregation reads the cumulative value for stored sessions too, with no change to your `main.py`. I also used the explicit-membership form (`"_cached_sum" in tok`) rather than `or`, so a legitimately-zero read can't be masked by the HWM.

**Verified** (throwaway DB, real numbers): a stored Claude session goes 80% → 99.5%; non-Claude agents (codex/gemini/…) come out byte-identical; and a simulated v1→v2 upgrade adds the column and degrades un-rescannable pruned rows to the old HWM rather than a false zero.

This stays your PR — you found the bug and shipped the core fix; I just completed the persistence side so every `/analytics` view is consistent. Appreciate the contribution.

PR #130, opened 2026-07-09
Session intent previews for IDE sessions + per-step token usage

VasiHemanth:
Implements the two ideas from discussions #128 and #129 (both from @ipatalas).

## #129 — Better session intent in sessions list

Sessions started from the VS Code / JetBrains Claude plugins prefix the first
user prompt with `<ide_selection>`/`<ide_opened_file>` context tags, and
sometimes ship content as block lists. The sessions list either showed the raw
tags or skipped the message entirely and fell through to "No prompt content".

- New `_claude_user_prompt_preview()` in `backend/main.py`: handles string and
  content-block-list messages, strips well-formed `<ide_*>…</ide_*>` and
  `<system-reminder>…` blocks, and skips meta lines / command echoes / tool
  results so the scan falls through to the next real user prompt.
- Preview length raised to 200 chars (was 120).
- Tests in `backend/test_prompt_preview.py` (10 cases, including unclosed-tag
  fallback and ide-context-only messages).

## #128 — Show token usage per step

Per-step usage existed only in the raw pane. The trace now shows it inline:

- A small chip under each step's cards: `⚡ 1.4k out · 4 in · 45.2k cache read`.
- A compact output-token count on each row of the step index rail (full
  breakdown in the tooltip), so the main offender stands out when scanning.

Data sources per agent:
- **Claude / Cursor**: `message.usage` per assistant line, deduped by message
  id (one API call spans several JSONL lines repeating the same usage).
- **Codex**: `token_count` event_msgs stay filtered from the trace, but their
  `last_token_usage` is attached to the preceding kept event first.
- **OpenCode**: `step-finish` parts now attach their tokens to the step's last
  trace event in the detail endpoint.

## Validation

- `backend/test_prompt_preview.py`: 10/10 pass; full backend suite has the
  same 17 environment-dependent failures as origin/main baseline (verified by
  stash/run/pop), no new failures.
- `frontend`: `tsc --noEmit` clean, `next build` succeeds.
- `UPDATE.json`: new 2026-07-09 release entry with both highlights.

### DIFF
--- a/UPDATE.json
+++ b/UPDATE.json
@@ -1,6 +1,21 @@
 {
   "releases": [
     {
+      "tag": "2026-07-09",
+      "title": "Readable session intents & per-step tokens",
+      "highlights": [
+        {
+          "title": "IDE sessions now show your actual prompt",
+          "description": "Sessions started from the VS Code or JetBrains Claude plugins used to show \"No prompt content\" or raw editor tags in the sessions list. The list now skips the editor context and shows the first words you actually typed, so you can find the right session at a glance."
+        },
+        {
+          "title": "Token usage on every step of a trace",
+          "description": "Open a session trace and each step now shows the tokens that API call used — a small chip under the step and a per-step count in the step index. Makes it easy to scan an expensive session and spot which step was the main offender.",
+          "href": "/sessions"
+        }
+      ]
+    },
+    {
       "tag": "2026-06-24",
       "title": "Git worktrees, grouped under their project",
       "highlights": [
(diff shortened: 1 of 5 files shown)


VasiHemanth:
Follow-up pushed (2f4b658) after navigating the live UI: the empty-intent rows weren't only IDE Claude sessions.

- **Copilot (VS Code)**: VS Code writes a `chatSessions` file the moment the chat panel opens — 44 of 135 local files had zero requests and showed as sessions with empty intent and 0 tokens. The scanner now skips files with no requests and no pending requests (136 → 92 copilot sessions locally, 0 empty intents).
- **OpenCode**: first-message previews could surface a raw `<system-reminder>` block. The tag-stripping now lives in `_strip_context_tags()` and applies to OpenCode text parts too, falling through to the next text part when nothing remains.

Verified in the running dashboard: project activity pages show no empty or tag-prefixed intents; backend suite unchanged apart from 3 new passing tests.

PR #131, opened 2026-07-10
fix(scan): remove 100-session cap, fix timestamp/cache correctness for Claude & Codex

mariadb-KyleHutchinson:
## Summary
Claude and Codex session scans stopped being trustworthy once a machine had more than 100 sessions: everything past the 100 most-recent was permanently stubbed at zero tokens, and reopening an old session could misdate it into "today." This branch removes the cap (backed by a new mtime-keyed sidecar cache so full scans stay fast), fixes timestamp/project/artifact/analytics correctness bugs the cap-removal exposed along the way, and hardens the new cache against a path-traversal input and a stub-vs-real data race.

### Features
- **scan:** add a reusable, mtime-keyed sidecar parse cache (`scan_cache.py`) so Claude/Codex session parsing can skip reparsing unchanged transcripts, with freshness tracked inside the JSON payload rather than relying on filesystem mtimes.

### Fixes
- **claude:** derive a session's reported timestamp from its last real user/assistant turn instead of the file's mtime — reopening an old session (e.g. via VS Code's history sidebar) no longer misdates it as happening today, which previously scattered old work into the wrong `/analytics` day bucket and could trip a false recent-activity flag.
- **claude:** remove the `[:100]` parse cap that silently froze every session past the 100 most recent as a zero-token stub; every discovered session is now parsed (or served from cache).
- **claude:** exclude `artifacts` from the sidecar cache payload so Claude Project Memory files are always rediscovered fresh, and stop `delegated_cost` from appearing as an explicit `null` on cache hits for delegation-less sessions.
- **claude:** keep MCP-tool and Skill usage analytics (`mcp_usage`/`skills_used`/`tool_counts`) intact on a cache hit — they were previously omitted from the cache payload, so the ecosystem-analytics aggregation silently undercounted toward zero as caches warmed up, with no error surfaced anywhere.
- **codex:** remove the equivalent `[:100]` parse cap and wire in the same sidecar cache, keyed on the max mtime across a session's (possibly multi-file, resumed/forked) rollout files.
- **codex:** re-run project aliasing from the cached raw cwd on every cache hit, so a session's reported project no longer freezes at `"unknown"` once its cache goes warm, and alias-table edits keep applying retroactively.
- **codex:** gate the stub→real flip on `source_mtime` (matching Claude) so an index-only session with no rollout file can no longer flip to "real" and zero-crush a previously-persisted row — the core hazard class this branch exists to fix.
- **history:** branch `upsert_sessions`'s `ON CONFLICT` clause on a `stub` flag so a stub row (partially known this scan) only touches liveness columns, never overwriting a previously-persisted real rollup's tokens/cost.
- **scan:** reject path-unsafe session ids in `scan_cache` (mirrors the existing subagent-trace guard).

### Other
- **refactor(scan):** seed a transient `stub` flag on every Claude/Codex session dict at construction, flipped to `False` once a full parse (or cache hit) completes.
- **test:** isolate `scan_cache` writes under `TOKENTELEMETRY_DATA_DIR` in the `scan_env` fixture so no scan test can leak cache files into a real `~/.tokentelemetry`; extensive new coverage for the cache, the stub-vs-real upsert guard, and the removed parse caps.
- **docs:** add an `UPDATE.json` entry (full session-history coverage + accurate session dating) per this repo's pre-push policy for the `feat:` commit above.

### DIFF
--- a/UPDATE.json
+++ b/UPDATE.json
@@ -6,7 +6,7 @@
       "highlights": [
         {
           "title": "Full session history — not just the last 100",
-          "description": "Claude Code and Codex used to stop tracking real usage once you passed 100 sessions; anything older sat frozen as a blank, zero-token row. Every session is now fully scanned, with a local cache keeping repeat scans fast, so your full history — and its true token/cost totals — stays accurate no matter how many sessions you've run.",
+          "description": "Claude Code and Codex used to stop tracking real usage once you passed 100 sessions; anything older sat frozen as a blank, zero-token row. Every session is now fully scanned, with a local cache keeping repeat scans fast, so your full history — and its true token/cost totals — stays accurate no matter how many sessions you've run. Thanks to mariadb-KyleHutchinson for reporting and fixing this (PR #131).",
           "href": "/sessions"
         },
         {
--- a/UPDATE.json
+++ b/UPDATE.json
@@ -1,6 +1,22 @@
 {
   "releases": [
     {
+      "tag": "2026-07-09",
+      "title": "Every session tracked, correctly dated",
+      "highlights": [
+        {
+          "title": "Full session history — not just the last 100",
+          "description": "Claude Code and Codex used to stop tracking real usage once you passed 100 sessions; anything older sat frozen as a blank, zero-token row. Every session is now fully scanned, with a local cache keeping repeat scans fast, so your full history — and its true token/cost totals — stays accurate no matter how many sessions you've run.",
+          "href": "/sessions"
+        },
+        {
+          "title": "Reopening an old session no longer misdates it",
+          "description": "Opening an old Claude session (e.g. from VS Code's history sidebar) used to bump its reported date to today, scattering old work into the wrong day on Analytics and sometimes tripping a false \"recent activity\" alert. Sessions are now dated by when the conversation actually happened.",
+          "href": "/analytics"
+        }
+      ]
+    },
+    {
       "tag": "2026-06-24",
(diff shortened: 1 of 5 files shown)


VasiHemanth:
Thanks for this — the cap removal and the stub-vs-real upsert guard fix a real data-integrity bug, and the test coverage made it easy to build on. I pushed one maintainer commit (28a039d) onto your branch with review follow-ups:

- **`CACHE_VERSION` stamped into every cache payload and checked on read.** `_mtime` detects transcript changes; the version detects *code* changes. Without it, a future parsing or pricing fix would never reach already-cached sessions — their transcripts unchanged, their numbers frozen under the old logic forever. A mismatch reads as a miss, so old entries reparse transparently. (Your own branch history shows the hazard: `mcp_usage` joined the payload mid-branch, and caches written by the earlier commits would have lacked it permanently.)
- **Claude freshness now keys on the max mtime across the parent transcript and its `subagents/agent-*.jsonl` files.** Delegation totals come from those separate files, so a background subagent finishing after the parent's last write used to stay undercounted while the cache was warm.
- **Atomic cache writes** (mkstemp + `os.replace`) so a crash mid-write can't leave a torn file.
- **`stub` stripped from `/sessions` responses** — it's persist plumbing, not API surface.

Five new tests cover the version key, the late-finishing subagent, and the stub strip. Full suite passes (the one `test_analytics_ecosystem_aggregates` failure is a pre-existing environment issue on main, unrelated).

PR #144, opened 2026-07-14
fix + docs: security & docs findings from Shiv's review

VasiHemanth:
Addresses Shivprakash Swami's review — 2 of 3 security findings fixed (the third refuted), plus both docs findings. (Path traversal in `session_id` was **verified and refuted**: Starlette's `[^/]+` path converter plus uvicorn's single decode-pass return a 404 before the glob ever runs; `/artifacts` needs its `resolve()`/`is_relative_to()` guard only because its `path` is a freeform *query* param, a different mechanism.)

## 1. Token redaction in access logs (`fix(backend)`)
Artifact ``/`<a>` loads authenticate via `?token=` (browsers can't set an `Authorization` header on native resource loads), so the remote-access token was written to uvicorn's default access log — persisted under systemd/journald/docker. A `logging.Filter` on the `uvicorn.access` logger now rewrites `token=<value>` → `token=<redacted>` before formatting. Loopback-default local usage has no token; only non-loopback `--host` is affected.

## 2. Supply-chain hardening (`chore(deps)`)
Closes the registry-compromise-between-installs window (users resolved a fresh dep tree on every first launch):
- Commit `package-lock.json` for root/frontend/website; `bin/cli.js` uses `npm ci` (fallback `npm install`).
- `backend/requirements.lock` generated via `uv --generate-hashes`; `bin/cli.js` installs with `--require-hashes` (fallback to `requirements.txt`); `>=` pins bounded.
- CI security-audit runs `npm ci` against the committed locks.
- `.github/dependabot.yml` (weekly npm/pip/actions, grouped minor+patch) so committed locks don't freeze users on stale vulnerable pins.

## 3. Docs (`docs`)
- README: `./start.sh` is the relaunch command — there's no global `tokentelemetry` command (Shiv stumbled on this after first install). New **Updating** section (`git pull && ./start.sh`, deps self-heal on launch), plus a Troubleshooting entry.
- `install.sh`: re-running the one-line installer over an existing clone now does `git pull --ff-only` instead of silently relaunching stale code (a trap Shiv's report surfaced indirectly), and prints where it landed + how to relaunch. `--ff-only` skips a dirty/diverged checkout rather than clobbering it.

## Verification
- Redaction filter tested through uvicorn's real `AccessFormatter` — `SECRET123` never leaks (token first/last param + no-token control). Token values (`crypto.randomBytes(24).toString('base64url')`) never contain `&`, space, or quotes, so the regex charset has no false negatives.
- Hash-pinned lock installs and imports cleanly under **Python 3.14.6** and 3.11.7; server verified running on 3.14.6.
- `node --check bin/cli.js`, `bash -n install.sh`, `ast.parse(main.py)`, dependabot/audit YAML all pass.

## Expected behavior change
`npm ci` requires `package-lock.json` to be consistent with `package.json` — editing `package.json` without regenerating the lock will make `npm ci` (and this CI gate) **error out** rather than silently resolve a new tree. That fail-fast is the point: it forces the lock to stay in sync. Workflow after any `package.json` edit: run `npm install` locally (updates the lock), commit both.

## Follow-up (not in this PR)
Dependabot's `pip` ecosystem tracks `requirements.txt`, not the non-standard `requirements.lock` extension — a small CI step to re-run `uv pip compile --generate-hashes` on `requirements.txt` changes would keep the lock in sync automatically.


VasiHemanth:
Thanks for bringing this up, @shivswami 🙏 — this PR ships fixes for two of your three findings:

- **Access-log token redaction** — a `logging.Filter` on `uvicorn.access` now rewrites `token=<redacted>` before the line is formatted, so the remote-access token no longer lands in logs under systemd/journald/docker.
- **Supply-chain** — committed lockfiles + `npm ci`, a hash-pinned `requirements.lock` installed with `--require-hashes`, and a Dependabot config so the pins don't go stale.

On finding #1 (path traversal in `session_id`): I dug in and it refuted — Starlette's `[^/]+` path converter plus uvicorn's single decode-pass return a 404 before the glob ever runs, so an encoded `../` can't reach the handler. `/artifacts` needs its `resolve()`/`is_relative_to()` guard only because its `path` is a freeform *query* param, which is a different mechanism. If you've got a request that actually reproduces it, I'd genuinely like to see the exact bytes — a raw-socket test with `..%2f..%2f` came back 404 for me. Either way, adding a cheap belt-and-suspenders guard is on the table if you think it's worth it.

Really appreciate the thorough writeup. 🌟

PR #172, opened 2026-07-18
feat: Docker/Podman compose support + GHCR CI

slmingol:
## Summary

- Run the backend and frontend as containers using Docker or Podman
- `Makefile` with colored help and targets for dev (`make up`) and prod (`make up-prod`)
- GitHub Actions CI builds and publishes `linux/amd64` + `linux/arm64` images to GHCR on every push to main

## What changed

**Container support**
- `backend/Dockerfile` — Python 3.12-slim, installs from `requirements.lock` with `--require-hashes`
- `frontend/Dockerfile` — 3-stage build (`--platform=$BUILDPLATFORM` on deps/builder so `npm ci` and Next.js compile run natively, not under QEMU); `NEXT_PUBLIC_API_PORT` baked in as a build arg
- `backend/.dockerignore` / `frontend/.dockerignore`
- `frontend/next.config.ts` — added `output: "standalone"` for the lean final stage
- `compose.yml` — host ports 13000 (frontend) / 18000 (backend) to coexist with other containers; `~/.claude` mounted read-only; named `tt_data` volume for persistent state
- `docker-compose.prod.yml` — production overlay that pulls pre-built GHCR images instead of building locally; `GHCR_OWNER` and `TT_IMAGE_TAG` vars
- `Makefile` — auto-detects podman vs docker; `make up` skips build if images already exist; `make up-prod` pulls from GHCR; colored `make` help

**CI/CD**
- `.github/workflows/container-publish.yml` — builds backend + frontend in parallel, pushes `latest` + short-SHA tags to `ghcr.io/<owner>/tokentelemetry-{backend,frontend}` on main; PRs build only (no push); GHA layer caching

**Housekeeping**
- Node.js 20 → 24 in `frontend/Dockerfile`, `security-audit.yml`, `deploy-website.yml`
- `UPDATE.json` — release entry for this feature

## Test plan

- [x] `make` shows colored help
- [x] `make build` builds both images (quiet one-liner, full log on failure)
- [x] `make up` skips rebuild when images exist; prints correct host URLs (13000 / 18000)
- [x] Frontend reachable at http://localhost:13000, backend at http://localhost:18000
- [x] `make down` / `make clean` work
- [x] `GHCR_OWNER=<user> make up-prod` pulls and starts successfully
- [x] CI workflow passes on PR; pushes to GHCR on merge to main

### DIFF
--- a/.github/workflows/container-publish.yml
+++ b/.github/workflows/container-publish.yml
@@ -1,0 +1,110 @@
+name: Build and publish container images
+
+on:
+  push:
+    branches: [main]
+    paths:
+      - "backend/**"
+      - "frontend/**"
+      - "compose.yml"
+      - ".github/workflows/container-publish.yml"
+  pull_request:
+    branches: [main]
+    paths:
+      - "backend/**"
+      - "frontend/**"
+      - "compose.yml"
+      - ".github/workflows/container-publish.yml"
+  workflow_dispatch:
+
+permissions:
+  contents: read
+  packages: write
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+env:
+  REGISTRY: ghcr.io
+
+jobs:
+  publish-backend:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v7
+
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v3
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
+      - name: Log in to GHCR
+        if: github.event_name != 'pull_request'
+        uses: docker/login-action@v3
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Extract metadata
+        id: meta
+        uses: docker/metadata-action@v5
+        with:
+          images: ${{ env.REGISTRY }}/${{ github.repository_owner }}/tokentelemetry-backend
+          tags: |
+            type=raw,value=latest,enable=${{ github.ref == 'refs/heads/main' }}
+            type=sha,format=short
+
+      - name: Build and push backend
+        uses: docker/build-push-action@v6
+        with:
+          context: ./backend
+          platforms: linux/amd64,linux/arm64
+          push: ${{ github.event_name != 'pull_request' }}
+          tags: ${{ steps.meta.outputs.tags }}
+          labels: ${{ steps.meta.outputs.labels }}
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
+
+  publish-frontend:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v7
+
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v3
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
+      - name: Log in to GHCR
+        if: github.event_name != 'pull_request'
+        uses: docker/login-action@v3
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Extract metadata
+        id: meta
+        uses: docker/metadata-action@v5
+        with:
+          images: ${{ env.REGISTRY }}/${{ github.repository_owner }}/tokentelemetry-frontend
+          tags: |
+            type=raw,value=latest,enable=${{ github.ref == 'refs/heads/main' }}
+            type=sha,format=short
+
+      - name: Build and push frontend
+        uses: docker/build-push-action@v6
+        with:
+          context: ./frontend
+          platforms: linux/amd64,linux/arm64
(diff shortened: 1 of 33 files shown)


VasiHemanth:
Hi @slmingol, thanks for this, it's a solid contribution and the build/CI side is well done (the `$BUILDPLATFORM` fix and the macOS-to-Linux path translation for the VS Code/Cursor mounts are both exactly right). I ran a verification pass against the diff and main before merging and found a handful of container-specific issues, roughly in priority order:

**1. Auth token gate is silently disabled (security)**
Token auto-generation for non-loopback hosts lives in `bin/cli.js` only. The container runs `python main.py` directly, so `TT_AUTH_TOKEN` is never set and `RemoteAuthMiddleware` passes every request through (`backend/main.py:228-232`). Combined with `ports: "${TT_API_PORT:-18000}:8000"` (publishes on `0.0.0.0`), the API is reachable unauthenticated from the LAN, which the native `--host 0.0.0.0` path deliberately prevents. Suggested fix: bind publishes to loopback by default (`127.0.0.1:${TT_API_PORT:-18000}:8000`) and document setting `TT_AUTH_TOKEN` for LAN use.

**2. No `TZ` passthrough, so day buckets shift (data correctness)**
The backend buckets days in server-local time (`_bucket_key`, `backend/main.py:7119`; budget windows at `6904-6971`). The container defaults to UTC, so for any non-UTC user, sessions near midnight land in the wrong day and weekly/monthly budget windows start at the wrong hour. Adding `TZ: ${TZ:-}` (or an `/etc/localtime` mount) to the backend service fixes it.

**3. `notifications.db` resets on every `make up`**
`backend/notifications.py:30` resolves its path from `TOKENTELEMETRY_HOME`, not `TOKENTELEMETRY_DATA_DIR`, so it lands on the container's writable layer instead of the `tt_data` volume. Since `up` targets run `_cleanup` (`compose down`) first, notification dedup state is lost on every start and budget alerts re-fire. Setting `TOKENTELEMETRY_HOME=/tt-data` alongside the existing var covers it (I'll fix the env inconsistency in `notifications.py` on our side separately).

**4. `TT_IMAGE_TAG`/`TT_API_PORT` override breaks the browser silently**
`NEXT_PUBLIC_API_PORT` is baked at build time and `make up` skips rebuild when images exist, so changing `TT_API_PORT` later leaves the browser calling a dead port with no error. For `up-prod` the GHCR image has 18000 baked in permanently. The compose.yml comment covers this, but the README's "Override with `PORT=` / `TT_API_PORT=`" line needs the rebuild caveat, and up-prod probably shouldn't advertise the override at all.

**5. Podman + SELinux (Fedora/RHEL)**
The Makefile auto-selects podman, but the `:ro` mounts have no `:z`/`:Z`, so on an SELinux-enforcing host the scanner gets EACCES and the app silently shows no sessions. `ro,z` on the agent mounts fixes it (harmless elsewhere).

**6. Windows: `${HOME}` is unset**
Compose substitutes an empty string, mounts an empty `/.claude`, and starts fine with zero sessions. Worth either a `${HOME:?HOME not set}` guard in compose or a README note that the container path needs WSL/git-bash.

**7. Coverage gaps in the commented mounts**
opencode (`~/.local/share/opencode`), `~/.pi`, `~/.qwen`, `~/.vibe`, and `~/.cursor` are scanned by the backend but have no mount entries at all, commented or otherwise. The ones you did include all have correct targets.

Happy to take some of these as follow-ups on our side (2, 3, and the loopback default are small). Also toying with generating the mount list host-side from the backend's own agent-dir registry so users don't hand-edit compose.yml; that shouldn't block this PR.

slmingol:
Thanks for the thorough review. Addressed in the latest commit:

**Security**
- Both ports now bind to `127.0.0.1` by default. Added a comment in `compose.yml` documenting the `0.0.0.0` override pattern and the need to set `TT_AUTH_TOKEN` for LAN use.

**Data correctness**
- Added `TZ: ${TZ:-}` to the backend service so day buckets and budget windows follow the host timezone.

**notifications.db**
- Added `TOKENTELEMETRY_HOME=/tt-data` alongside `TOKENTELEMETRY_DATA_DIR` so `notifications.py` resolves its DB path to the persistent volume.

**SELinux**
- All `:ro` mounts updated to `:ro,z`; `${HOME}/.claude` now quoted. Added a header comment explaining the `:z` suffix.

**Missing agent mounts**
- Added commented entries for `cursor`, `vibe`, `qwen`, `opencode` (`~/.local/share/opencode`), and `pi`.

**Makefile podman detection**
- Changed from `command -v podman` to `podman compose version` so the fallback to `docker compose` fires correctly when `podman` exists but the compose subcommand is missing.

**README**
- Clarified that `TT_API_PORT` requires a frontend rebuild (baked into the Next.js bundle at image build time).
- Added a Windows note: WSL or Git Bash required — `${HOME}` is unset in `cmd.exe`/PowerShell.
- Noted that `TT_API_PORT` should not be overridden with `make up-prod` since GHCR images have the port baked in.

**`service_started` vs `service_healthy`**
- Kept `service_started` intentionally — `service_healthy` causes Podman's gvproxy to double-register port forwards on macOS, producing "proxy already running" errors on every `make up`. Added an inline comment explaining this.

Happy to take items 2 and 3 (TZ, notifications.db) as follow-ups on your side if you prefer to own those consistently with how the rest of the env vars are wired — they're in this commit but easy to revert and re-land from your side.

PR #173, opened 2026-07-18
feat(projects): recurring loops in a project's Config & Insights tabs

VasiHemanth:
Surfaces `/loop` automations at the **project** level — the Config tab lists the loops set up in a project, the Insights tab breaks down their token/cost economics.

## Stacked on #169
This builds on the loop `loop` field that **#169** adds to every session. **Merge #169 first, then this.** The PR is based on `feat/loop-telemetry-rebased` (the #169 head) so the diff below is only this feature's changes; retarget it to `main` once #169 lands.

## What's here
- **`_lib/loops.ts`** — a pure client-side fold over the project's session rows into the same `loops`/`by_loop` shape the `/analytics` endpoint produces. No backend change, no extra fetch: the scan already annotates each session's `loop` field (lifecycle state + fire-turn footprint), and the project tabs already hold the project-scoped session list. Footprint = the loop's own fire-response turns, keyed by `job_id || session_id`.
- **Config tab** — a *Recurring loops* inventory: every loop set up in the project with its state, interval, fire count, mode and job id, each card linking to its session trace. Shown as soon as the project opens; self-hides when there are none.
- **Insights tab** — a *Recurring loops* breakdown: active / expired / cancelled tiles, observed fire counts, and the token & cost footprint of loop turns in context of the whole session (`of $X session`).

## Verification
- `tsc --noEmit` clean.
- Client fold cross-checked against `/analytics?projects=<path>` for a project with 4 loops — **exact match**: total 4, expired 2, cancelled 2, ≥75 iterations, 805,100 loop tokens, \$85.995473 loop cost.
- Both routes render 200 with real data (screenshots captured live against the loop backend).

Loop tokens/cost stay an **attribution view** — already part of the session totals, never re-summed into them.

https://claude.ai/code/session_01UDo8qcPJDhxphuy4Cfskrn

### DIFF
--- a/UPDATE.json
+++ b/UPDATE.json
@@ -12,6 +12,10 @@
           "title": "Recurring loops in Analytics",
           "description": "A new Recurring loops section counts your active, expired and cancelled loops with their observed fire counts, cadence and cost, so a forgotten hourly loop quietly burning tokens is easy to spot. It's detected from the actual scheduling calls, so it catches loops even when the /loop command left no trace.",
           "href": "/analytics"
+        },
+        {
+          "title": "Per-project loop view",
+          "description": "Open any project and its Config tab now lists every recurring loop set up there — state, cadence, fire count and job id — so you can see at a glance which automations are still running in that repo. The Insights tab adds the numbers: active/expired/cancelled counts and the token & cost footprint of the loops' own fire turns, scoped to that project."
         }
       ]
     },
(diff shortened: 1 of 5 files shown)


VasiHemanth:
Superseded by #177, which consolidates this (project loop tabs) with #174 onto a clean branch off main. #169 was squash-merged, so this branch re-carried its loop commits and couldn't retarget to main without conflicts; #177 has only the net-new work.

PR #175, opened 2026-07-18
feat(settings): show each agent's experimental / feature flags

VasiHemanth:
Adds a read-only **Agent feature flags** section to Settings — at a glance, which experimental/preview features each agent has enabled in its own local config. Answers "is Copilot's experimental mode on?" without digging through dotfiles.

## Why
There is **no shared "experimental mode"** across coding agents — each exposes it differently (a boolean, a feature table, individual toggles) or server-side, or not at all. This surfaces the three with a clean, secret-free, locally-readable signal:

| Agent | Source | Flag(s) |
|---|---|---|
| **Copilot** | `~/.copilot/settings.json` | `experimental` — the boolean that unlocks `/every`, `/loop`, `/after` scheduled prompts |
| **Codex** | `~/.codex/config.toml [features]` | per-feature booleans (`js_repl`, …) |
| **Claude Code** | `~/.claude/settings.json` | `enableWorkflows`, `effortLevel`, `editorMode`, `voiceEnabled`, `agentPushNotifEnabled` |

## Design
- **Backend** `GET /config/agent-features` + `_agent_feature_flags()` — best-effort, read-only. A missing/corrupt config is omitted (never raised), and only an **allowlist of scalar, non-secret keys** is surfaced — no auth tokens, no whole-file dumps. An absent agent means *"not exposed locally"*, never *"off"*; server-side/opaque-store agents (Cursor, cloud harnesses) aren't listed rather than faked (same honesty rule as the billing/loop views).
- **Frontend** `AgentFeatureFlags` — per-agent card with on/off pills (green ✓ / muted ✗), the config path, and a one-line note.
- Local-first, no new outbound calls.

## Verification
Live against real config: Copilot `experimental` **on**, Codex `js_repl` **off**, Claude `enableWorkflows` on / `effortLevel` high / … (screenshot captured). `tsc --noEmit` clean. Independent of the loop PRs — branches off `main`, merges on its own.

https://claude.ai/code/session_01UDo8qcPJDhxphuy4Cfskrn

### DIFF
--- a/UPDATE.json
+++ b/UPDATE.json
@@ -1,6 +1,17 @@
 {
   "releases": [
     {
+      "tag": "2026-07-18",
+      "title": "See which experimental features each agent has on",
+      "highlights": [
+        {
+          "title": "Agent feature flags in Settings",
+          "description": "Settings now shows the experimental and preview features each coding agent has enabled in its own local config — at a glance whether Copilot's experimental mode (the one that unlocks /every and /loop), Codex's per-feature flags, or Claude Code's feature toggles are on or off. It's read-only and local-first; agents that gate features server-side aren't listed rather than faked.",
+          "href": "/settings"
+        }
+      ]
+    },
+    {
       "tag": "2026-07-16",
       "title": "See which sessions ran a loop",
       "highlights": [
--- a/UPDATE.json
+++ b/UPDATE.json
@@ -6,7 +6,7 @@
       "highlights": [
         {
           "title": "Agent feature flags in Settings",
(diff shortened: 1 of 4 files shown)


VasiHemanth:
### Update: guide users to enable it, don't write their config

Considered making the pills editable (toggle the flag from TT), but decided against it: agents like **Copilot restart** when the flag changes, most only read these at startup, and writing another tool's config file is outside TT's read-only observer role.

Instead each card now shows **how to turn it on in the agent itself** — the exact command as a chip (`/experimental`, Codex `[features]`, `/config`) plus a **Docs** link. All three doc URLs were fetched and verified live:
- Copilot → `docs.github.com/.../copilot-cli/automate-copilot-cli/schedule-prompts`
- Codex → `learn.chatgpt.com/docs/config-file/config-reference` (the `developers.openai.com/codex/config-reference` redirect target)
- Claude Code → `code.claude.com/docs/en/settings`

Still read-only, local-first. Verified live in the browser.

PR #180, opened 2026-07-18
feat(org): self-hosted org mode MVP — collector, shipper, Team page

VasiHemanth:
## What this is

First slice of org mode: teams run their own central collector, each dev machine ships metric rollups to it, and a new Team page shows the aggregate. Nothing changes for single-user local use; without an org.json the endpoints report disabled and the page shows setup instructions.

## Pieces

- **`POST /org/ingest`** — accepts session rollups (id, agent, project, timestamp, token counts, optional model/cost). Auth is a per-machine token (`X-TT-Machine-Token`) resolved against `<data_dir>/org.json`, compared constant-time against every configured token. The endpoint is exempt from the `TT_AUTH_TOKEN` dashboard gate (remote agents don't hold that token) but its own machine-token check is mandatory even from loopback. Batch capped at 1000 sessions; string fields length-capped; malformed bodies 422 before any write.
- **`backend/org_store.py`** — SQLite at `<data_dir>/org.db`, PRIMARY KEY `(machine, session_id)`, upsert returns (accepted, updated) so re-scans never double-count. `summary()` rolls up totals, by-machine/by-agent/by-project, and the 20 most recent sessions ordered by `datetime(timestamp)` so mixed UTC offsets sort by real instant.
- **`backend/org_ship.py`** — stdlib-only shipper for cron/launchd: reads the local backend's `/sessions`, maps to rollups, posts in batches. `--central-url` / `--token` flags or `TT_CENTRAL_URL` / `TT_MACHINE_TOKEN` envs.
- **Frontend `/team`** — totals cards, per-machine (with last-seen), per-agent and per-project tables, recent sessions; empty state walks through collector + shipper setup. Nav entry added.
- **`backend/test_org_mode.py`** — 14 tests: valid/invalid/missing token, idempotent re-send, re-send with changed values must overwrite (kills a DO NOTHING regression), mixed-offset recent ordering, non-ASCII token answers 401 not 500, oversize batch 422, body-label spoof ignored, corrupt org.json treated as disabled, remote-client ingest with TT_AUTH_TOKEN set.

## Verification

- `pytest backend/test_org_mode.py backend/test_remote_auth.py` — 27 passed.
- `tsc --noEmit` in `frontend/` — clean.
- Independent recompute: seeded two machines with known counts via `org_store` directly; hand-computed totals (3 sessions / 600 tokens / $2.00), per-machine and per-agent splits, and instant-ordering of `recent` across `+05:30`/`+00:00` offsets all match `summary()` exactly. Re-ingest of an identical batch returned accepted 0 / updated 2 with totals unchanged.
- Built with a multi-agent workflow; three adversarial review passes (auth, data correctness, FE/BE contract) surfaced four real issues — unbounded ingest batch, non-ASCII token crash, lexical timestamp ordering, and an upsert-blind test suite — all fixed and covered by new tests.

## Not in this PR (deliberate)

Docker Compose packaging for the collector, machine-token management CLI, cost/model enrichment in the shipper, per-team grouping, and any managed-cloud/multi-tenant layer.

https://claude.ai/code/session_019aPyCUWq54vgUopuZnZww7

### DIFF
--- a/UPDATE.json
+++ b/UPDATE.json
@@ -2,6 +2,21 @@
   "releases": [
     {
       "tag": "2026-07-18",
+      "title": "Org mode: see your whole team's usage in one place",
+      "highlights": [
+        {
+          "title": "Team dashboard (early preview)",
+          "description": "A new Team page rolls up sessions, tokens and cost across every machine that ships metrics to your own self-hosted collector — split by machine, agent and project. Only metric rollups travel; prompts, code and file paths never leave a dev's machine.",
+          "href": "/team"
+        },
+        {
+          "title": "Ship metrics from each dev machine",
+          "description": "A small stdlib-only shipper (backend/org_ship.py) reads your local session list and posts rollups to the collector on a schedule, authenticated by a per-machine token. Re-sends are idempotent, so nothing double-counts. Setup steps live on the Team page until data arrives."
+        }
+      ]
+    },
+    {
+      "tag": "2026-07-18",
       "title": "See which experimental features each agent has on",
       "highlights": [
         {
--- a/backend/main.py
+++ b/backend/main.py
@@ -9,7 +9,7 @@ import yaml
 import sqlite3
 from pathlib import Path
 from typing import List, Optional, Dict, Any, Set, Iterable, Tuple
-from pydantic import BaseModel
+from pydantic import BaseModel, Field
 from datetime import datetime, timezone, timedelta, time as _dtime
 from urllib.parse import unquote, quote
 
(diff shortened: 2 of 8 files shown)


VasiHemanth:
@mariadb-KyleHutchinson this is the first cut of the org mode we talked about, and since the idea came from your questions, want to kick the tires before it merges?

What's in it: a central collector (the normal backend with an `org.json` next to it), a small shipper each dev machine runs on a schedule, and a Team page that rolls up sessions, tokens and cost by machine, agent and project. Only metric rollups travel; prompts, code and file paths never leave a machine.

To try it on one machine (shipping your own data to yourself):

1. `gh pr checkout 180`
2. Create `org.json` in your TokenTelemetry data dir (`~/.tokentelemetry/` by default):
   ```json
   {"machines": [{"label": "kyle-laptop", "token": "test123"}]}
   ```
3. Start the app as usual, then run:
   ```
   python backend/org_ship.py --central-url http://127.0.0.1:8000 --token test123
   ```
4. Open the Team page in the sidebar.

Re-running the shipper is safe; ingest is idempotent per machine and session, so nothing double-counts. With a second machine you'd add another entry to `org.json` and point its shipper at the collector's URL.

Things I'd especially like your read on: does the metrics-only payload pass the sniff test for what your org could run, and does the ingest schema look sane as a base (the store is SQLite for now, but the shape is what a MariaDB backend would inherit later). Any feedback as a review comment here is great, or a PR against this branch if you'd rather show it in code.

PR #182, opened 2026-07-19
fix: Codex Desktop session crash on structured reasoning summaries (#181)

### LINKED ISSUE #181: [BUG] Codex Desktop session crashes when rendering structured reasoning summaries
## Describe the bug
Opening certain Codex Desktop sessions causes the session detail page to crash with a React runtime error:

`Objects are not valid as a React child (found: object with keys {type, text})`

The affected Codex session contains reasoning events whose `payload.summary` is an array of structured objects such as:

```json
[
  {
    "type": "summary_text",
    "text": "Planning custom HTTP client logging handler"
  }
]
```

The session renderer appears to pass these objects directly to React instead of extracting their `text` values.

## To Reproduce

1. Run `./start.sh`.
2. Open TokenTelemetry at `http://localhost:3000/`.
3. Navigate to a Codex Desktop session containing structured reasoning summaries:
   `http://localhost:3000/sessions/[SESSION_ID]?agent=codex&from=%2F`
4. Observe that the session detail page crashes with a React runtime error.

## Expected behavior
The session detail page should load successfully and render the text from structured reasoning summaries. Unexpected or unsupported content shapes should be handled gracefully rather than crashing the entire page.

## Environment
- **OS**: macOS 26.5.2 (Darwin 25.5.0)
- **Node.js version**: v24.18.0
- **Python version**: Python 3.9.6
- **Agent(s) in use**: ChatGPT/Codex Desktop with GPT 5.6 Sol / Luna
- **Codex Desktop version**: 26.715.31925
- **TokenTelemetry version**: 1.0.0, commit `7fb578c`
- **Next.js version**: 16.2.10 (Turbopack)

## Logs / Screenshots
```text
## Error Type
Runtime Error

## Error Message
Objects are not valid as a React child (found: object with keys {type, text}). If you meant to render a collection of children, use an array instead.

    at SessionDetailPage/<.children<.children<.children<.children<.children< (src/app/sessions/[id]/page.tsx:795:29)
    at SessionDetailPage (src/app/sessions/[id]/page.tsx:775:35)

## Code Frame
  793 | return (
  794 |   <div key={idx} ref={(el) => { stepRefs.current[idx] = el; }} ...>
> 795 |     <EventCard
      |     ^
  796 |       event={event}
  797 |       mode={splitView ? "dialogue" : "all"}
  798 |       agent={agent}
  799 |       tokens={stepTokens[idx]}
  800 |     />
  801 |   </div>
  802 | );

Next.js version: 16.2.10 (Turbopack)
```

## Additional context
The failure appears to originate in the Codex `response_item` reasoning branch of `EventCard`.

When `payload.content` is absent, the renderer falls back to `payload.summary`. Newer Codex Desktop session data can represent `payload.summary` as an array of `{type, text}` objects. That array is currently used as a React child without first converting it to a string.

The renderer should normalize both string and structured-array forms, for example by extracting and joining each summary item's `text` field. A regression test using a Codex reasoning event with a structured `summary` array would help prevent this from recurring.

VasiHemanth:
## What

Fixes #181 — opening a Codex Desktop session with structured reasoning summaries crashed the entire session detail page with:

> Objects are not valid as a React child (found: object with keys {type, text})

## Cause

Newer Codex Desktop reasoning events carry `payload.summary` as an array of `{ type, text }` objects. In `EventCard`'s `response_item` reasoning branch, when `payload.content` was absent the code fell back to `payload.summary` and passed that raw array of objects straight to React as a child, crashing the page.

## Fix

- Added `coerceReasoningText` (`frontend/src/lib/reasoning.ts`) — flattens any reasoning value shape (plain string, array of `{type,text}` summary objects, nested object) into a display string. Unknown/empty shapes degrade to `""` instead of throwing.
- Routed the Codex reasoning branch through it, replacing the ad-hoc inline extraction.
- Added `reasoning.test.ts` (`node:test`, same pattern as `navigation.test.ts`) covering the reported array-of-summaries shape, single/mixed arrays, nested objects, alternate keys, and graceful degradation.

## Verification

- `node --test` on the new suite: 7/7 pass (exercised against the real module).
- `tsc --noEmit`: clean.

### DIFF
--- a/frontend/src/app/sessions/[id]/page.tsx
+++ b/frontend/src/app/sessions/[id]/page.tsx
@@ -16,6 +16,7 @@ import SummaryPanel from "@/components/summarizer/SummaryPanel";
 import { apiFetch, artifactUrl } from "@/lib/api";
 import { formatTokens, formatCost } from "@/lib/format";
 import { resolveSessionBackTarget } from "@/lib/navigation";
+import { coerceReasoningText } from "@/lib/reasoning";
 
 interface Artifact {
   name: string;
@@ -2007,9 +2008,7 @@ function EventCard({ event, mode = "all", agent, tokens }: { event: any, mode?: 
               {renderTimestamp()}
             </div>
             <div className="text-[var(--tt-fg-muted)] whitespace-pre-wrap italic text-xs leading-relaxed font-mono opacity-80">{
-              Array.isArray(payload.content)
-                ? payload.content.map((c: any) => c?.text ?? c?.summary ?? c?.content ?? (typeof c === 'string' ? c : "")).filter(Boolean).join("\n\n")
-                : (typeof payload.content === 'string' ? payload.content : (payload.summary ?? payload.text ?? ""))
+              coerceReasoningText(payload.content ?? payload.summary ?? payload.text)
             }</div>
           </div>
        );
--- a/frontend/src/lib/reasoning.test.ts
+++ b/frontend/src/lib/reasoning.test.ts
@@ -1,0 +1,54 @@
+import { test } from "node:test";
+import assert from "node:assert/strict";
+import { coerceReasoningText } from "./reasoning";
+
+test("plain string passes through", () => {
+  assert.equal(coerceReasoningText("Planning the client"), "Planning the client");
+});
+
+// Regression for issue #181: Codex Desktop reasoning summaries arrive as an
+// array of { type, text } objects. Passing that array to React as a child
+// crashes the whole session detail page ("Objects are not valid as a React
+// child"). It must be flattened to the joined text instead.
+test("structured Codex summary array is flattened to text", () => {
+  const summary = [
+    { type: "summary_text", text: "Planning custom HTTP client logging handler" },
+    { type: "summary_text", text: "Wiring the retry path" },
+  ];
+  assert.equal(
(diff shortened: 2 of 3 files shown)


VasiHemanth:
Thanks @vipentti for the precise report and root-cause pointer in #181 — the structured `summary` array shape and the suggested `text`-join fix were exactly right.

Validated live against a real Codex Desktop session, not just unit tests.

**Repro (pre-fix, current `feat/local-model-insights` on :3000):** opened session `019cf142-…-4a31`, which has 168 reasoning items in the plaintext `summary: [{type:"summary_text", text:…}]` shape. Reproduces the exact issue error:

> Objects are not valid as a React child (found: object with keys {type, text})

**After fix (this branch on :3001, same backend):** the same session loads fully (2467 steps / 750 reasoning events), console is clean of the React-child error, and the Reasoning cards render the extracted summary text (e.g. "Planning codebase verification …").

**Also:** a fresh `codex exec` run ingested and rendered fine. Worth noting Codex **CLI** reasoning summaries are encrypted (`reasoning summaries: none`), so only Codex **Desktop** sessions carry the plaintext array that triggered this — which is exactly what the fixture-free real repro used.

Checks: `node --test` reasoning suite 7/7, `tsc --noEmit` clean. Ready to merge.

PR #197, opened 2026-07-25
docs: plan /goal telemetry across the four agents that ship it

VasiHemanth:
Planning doc for `/goal` support, the companion to the `/loop` telemetry already shipped. No code changes; this is the design to build against.

## What the research found

Four of roughly fourteen supported agents ship `/goal` natively, and all four mean something different by it. Checked two ways for every agent, local artifacts **and** documentation, because local emptiness proves nothing: Codex shipped Goal Mode to GA in CLI 0.133 on 2026-05-21 while one of its two goal DBs on this machine sat empty.

| Agent | Mechanism | Durable evidence | Local usage |
|---|---|---|---|
| Codex | `thread_goals` SQLite with native `tokens_used` / `time_used_seconds` / `status` | dedicated DB | 1 goal: **270,359 tokens over 44 min**, status `paused` |
| Claude Code | session-scoped Stop hook + evaluator | transcript text only | 15 sessions armed, 30 goals, 15 real blocks (all in one session) |
| Grok Build | `update_goal` progress checkpoints with a `completed` flag | `chat_history.jsonl` tool calls | 88 calls / 24 sessions |
| Antigravity | prompt marker only | `transcript.jsonl` | ~19 invocations / 10 conversations |
| Cursor, Copilot, OpenCode | not native; third-party `cursor-goal` port writes `~/.cursor-goal/data/goal.json` | only if installed | not installed here |
| Gemini CLI, Qwen, Vibe, Pi, Cline, Hermes, Omnigent | none found in docs or data | — | — |

## The main design call

`/loop` got away with one shared schema because a recurring timer really is the same object in every agent. Goals are not, so this plan uses a shared **envelope** plus per-agent **evidence**, with:

- **`state_source: "reported" | "inferred"`** separating statuses the agent actually wrote down (only Codex ever does) from ones we guessed.
- **Per-agent allowed-state sets**, enforced in the annotator, so Claude can never emit `complete`. It has **no terminal event at all**: every hook record is either "Goal set" or "Stop hook feedback". 13 of 15 goal sessions recorded zero blocks, so for almost every goal the honest reading is "armed, outcome unknown".
- **`goals` as a list**, not a singleton like `loop`. Sessions arm goals repeatedly (one armed five).
- **Three separate cost bases that must never be summed.** The local Codex goal is the clearest case: its 270,359 tokens are ~5% *of* its session's 5,664,120, not an extra 270k on top.

## Detection traps recorded

- **Sessions that discuss goals look like sessions that ran them.** This research session alone generated 2 phantom blocks, 12% of the naive machine-wide total.
- Compaction replays duplicate the arm record at a second line number with an identical timestamp.
- "Tool present in the toolset" is not "tool was called": 1,808 candidate Grok files reduced to 88 real invocations.
- Codex ships two copies of its goals DB **at different migration levels**, and the populated one is on the *older* schema without the deferrals table.
- The 8-consecutive-block cap is real but comes only from the changelog (not the public hooks docs) and is overridable via `CLAUDE_CODE_STOP_HOOK_BLOCK_CAP`.

## Phasing

Codex first: the goal-to-session join was verified end to end against the running API (`backend/main.py:4411` rebuilds the thread UUID, and `thread_goals.thread_id == sess["id"]` needs no translation layer), so it needs a reader plus a card and no inference. Then Claude (most sessions, least evidence), then Grok and Antigravity. Roughly two days total.

### DIFF
--- a/docs/design/README.md
+++ b/docs/design/README.md
@@ -13,3 +13,4 @@ it in the feature's PR.
 | Durable history + analytics date filters | [durable-history.md](durable-history.md) | [ADR-0002](../adr/0002-durable-history-rollup.md) |
 | Delegation & ecosystem telemetry | [`../../DESIGN.md`](../../DESIGN.md) | — (pre-ADR) |
 | Documentation site + community resources | [documentation-site.md](documentation-site.md) | [ADR-0003](../adr/0003-docs-site-fumadocs.md) |
+| `/goal` telemetry (plan) | [goal-telemetry.md](goal-telemetry.md) | — |
--- a/docs/design/goal-telemetry.md
+++ b/docs/design/goal-telemetry.md
@@ -1,0 +1,317 @@
+# Design: `/goal` telemetry
+
+Companion to the `/loop` support already in `backend/main.py`
+(`_annotate_loop_lifecycle`, `_grok_loop_detect`, `_cline_loop_specs`). Same
+question, different feature: when a user hands an agent an objective and walks
+away, what did that cost and how did it end?
+
+Status: plan. Nothing here is implemented yet.
+
+## What `/goal` actually is
+
+`/goal` is not one feature. Four agents ship a command with that name and all
+four mean something different by it, with very different amounts of evidence
+left on disk. Everything below was verified against local docs bundles, the
+Claude Code changelog, and this machine's own session data (read-only).
+
+### Claude Code
+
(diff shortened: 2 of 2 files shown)


VasiHemanth:
Superseded: folding the design doc and the full implementation (all agents, not just Codex) into a single PR.

PR #201, opened 2026-07-26
feat(goals): surface Codex Goal Mode on the session trace

VasiHemanth:
Phase 1 of [`docs/design/goal-telemetry.md`](https://github.com/VasiHemanth/tokentelemetry/pull/197) (design PR #197). Codex is the only supported agent that keeps first-class `/goal` state on disk **and** counts a goal's tokens and wall-clock itself, so it ships first and needs no inference.

Before this, a Codex session that spent 270k tokens under an unattended goal rendered as an ordinary session.

## What it does

`backend/codex_goals.py` reads `~/.codex/**/goals_1.sqlite` and returns `thread_id -> [goal]`. `thread_id` is the Codex session id verbatim (`main.py` rebuilds the same UUID from the rollout filename), so the join needs no translation layer.

The card shows the objective, Codex's own status, tokens used, time in goal, token budget, and the goal's **share of the session**.

## Three things this deliberately gets right

- **Goals attach after the per-session cache branch.** Status is live mutable state (`active → paused → complete`); caching it would freeze whichever status we saw first, exactly the trap `_annotate_loop_lifecycle` exists to avoid.
- **Codex ships two goal DBs at different migration levels, and the populated one is on the *older* schema.** So the DB is picked by row count rather than by path, and `sqlite_master` is probed before touching the v2-only deferrals table. A v1 DB reports `deferrals: null`, never a confident `0`.
- **An unrecognised status degrades to `unknown`**, not to a plausible-looking `active`, with the agent's raw string kept in `evidence.status_raw`.

Goal tokens are shown as a share (4.8% locally) because Codex counts them *inside* the session total. Presenting them as incremental spend would double-count.

## Verification

Not just unit tests. Three independent paths agree on **270,359 tokens / 2,615 s / `paused`**:

| Source | Result |
|---|---|
| `sqlite3` CLI, straight at the DB | `019ebfe9…\|paused\|270359\|2615` |
| `GET /sessions` | same, `state_source: reported` |
| Rendered card | `270,359` · `43m 35s` · `4.8%` |

The 4.77% share recomputes independently from the endpoint's session total, and the timestamps round-trip correctly through IST (`07:38:03Z` → `1:08:03 PM`).

- **27 new tests**, and the suite was **mutation-checked**: first-db-wins, `unknown→active`, and `null→0` deferrals each make it fail.
- Full backend suite: **342 passed** vs **315 on `origin/main`**, with the identical 23 pre-existing failures (verified against a clean `origin/main` worktree).
- `tsc --noEmit` clean.

## Not in this PR

Claude Code, Grok and Antigravity goals (phases 2 and 3). Those need inference and a weaker honesty model; the design doc covers why.

### DIFF
--- a/UPDATE.json
+++ b/UPDATE.json
@@ -1,6 +1,21 @@
 {
   "releases": [
     {
+      "tag": "2026-07-26",
+      "title": "See what Codex Goal Mode actually cost",
+      "highlights": [
+        {
+          "title": "Goal Mode on the session trace",
+          "description": "Sessions where you ran Codex's /goal now show a Goal Mode card: the objective, Codex's own status (active, paused, complete or blocked), tokens used, time spent in the goal, and any token budget. A long unattended goal is no longer indistinguishable from an ordinary session.",
+          "href": "/sessions"
+        },
+        {
+          "title": "Goal cost shown as a share, not a second bill",
+          "description": "The card reports a goal's tokens as a percentage of its session total, because Codex counts them inside that total. The numbers come from Codex itself rather than being estimated, and status is re-read on every load so a finished goal never keeps showing as running."
+        }
+      ]
+    },
+    {
       "tag": "2026-07-23",
       "title": "Active loops now show their next run",
       "highlights": [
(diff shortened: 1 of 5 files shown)


VasiHemanth:
Superseded: rolling phase 1 into one PR that implements the whole design.

PR #219, opened 2026-08-01
feat: endpoint presets for the OpenAI-compatible summarizer

VasiHemanth:
Alternative to #218, which adds a dedicated MiniMax adapter.

MiniMax speaks the OpenAI chat-completions API, so it already works through the
`openai_compat` backend. The only friction was knowing the base URL and model id.
This adds preset buttons that fill both in.

## Changes

- `ENDPOINT_PRESETS` in `frontend/src/lib/summarizer.ts` — MiniMax, MiniMax (China),
  local llama.cpp. Each entry is an endpoint plus a starting model id.
- Preset buttons above the Endpoint field in the existing `OpenAICompatForm`
  (`BackendPicker.tsx`). Clicking one fills endpoint + model; the API key and
  sampling knobs stay the user's.

Two files, ~35 lines. Adding the next provider is one array entry.

## Why not a per-vendor adapter

Verified against the live endpoint and MiniMax's docs:

- An unauthenticated POST through `OpenAICompatSummarizer` to
  `https://api.minimax.io/v1/chat/completions` returns MiniMax's 401 body, and
  `errors.classify()` already buckets it as `auth` with the correct hint.
- On the OpenAI route with `reasoning_split` unset (the default), M3 inlines its
  thinking into `content` wrapped in `<think>` tags. `openai_compat` already
  strips those (`_THINK_RE`, `openai_compat.py:53`). #218's `_extract_openai_text`
  does not, so it would hand the full chain-of-thought to `parse_narrative`.
- #218 pins a hardcoded `SUPPORTED_MODELS` tuple and `save_config` silently
  rewrites the model back to the default when it isn't in that list. A new MiniMax
  model would be unusable until we cut a release. The free-text model field has no
  such gate.

## Checks

- `npx tsc --noEmit` in `frontend/` — clean
- `python3 -c "import json; json.load(open('UPDATE.json'))"` — parses
- Live 401 probe through the existing adapter (above)

Not yet done: an authenticated end-to-end summarize. Marking draft until that runs.

### DIFF
--- a/UPDATE.json
+++ b/UPDATE.json
@@ -1,6 +1,17 @@
 {
   "releases": [
     {
+      "tag": "2026-08-01",
+      "title": "One-click endpoints for hosted summarizers",
+      "highlights": [
+        {
+          "title": "Endpoint presets in the OpenAI-compatible backend",
+          "description": "The summarizer's OpenAI-compatible backend now has preset buttons that fill in the endpoint and a starting model for you. MiniMax (global and China) and a local llama.cpp server are there; paste your API key and hit Test connection. Hosted providers that speak the OpenAI API work through this backend, so there's no separate setup per vendor.",
+          "href": "/settings"
+        }
+      ]
+    },
+    {
       "tag": "2026-07-30",
       "title": "See what /goal runs actually cost",
       "highlights": [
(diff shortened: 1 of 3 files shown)


VasiHemanth:
Superseded — folding this into #218 instead of running a parallel PR. Details there.

PR #250, opened 2026-08-07
fix(cost): fold delegated cost into by_day/by_model, price new Claude models, add cost-by-day table

mariadb-KyleHutchinson:
## Summary
This branch originally fixed a duplicate-JSONL-line/cache-read undercount in Claude session scanning. #248 landed an equivalent, independently-authored fix (`message.id` dedup + cumulative cache-read billing via a dual `cached`/`_cached_sum` design) while this branch was in flight, so it's now rebased on top of that and only carries what #248 didn't already cover: delegated (subagent/workflow) cost was still missing from `by_day`, `by_model`, and the `/analytics` grand total; four new Claude models had no pricing entry and silently fell back to the generic default rate; and historical (non-today) `/analytics` queries never persisted delegated cost at all. Also adds a cost-by-day breakdown table to the Analytics page.

### Fixes
- **cost:** Fold delegated (Claude subagent/workflow) cost and tokens into `by_day`/`by_model`/total in `/analytics` (`by_agent` already had this) — subagent spend is attributed to its own model, which can differ from the parent's (e.g. Explore on Haiku under an Opus session). Add curated pricing for `claude-sonnet-5` (introductory rate through 2026-08-31), `claude-opus-5`, `claude-opus-4-8`, `claude-fable-5`, verified against Anthropic's published rate card rather than the models.dev overlay. Log a one-time warning when a model falls through to the generic `_default` rate, so the next new-model release degrades visibly instead of silently undercounting for weeks. Bumps the sidecar cache version (7→8) since the cached payload shape gained a new `delegated_by_model` field.
- **history:** Persist `delegated_cost`/`delegated_input`/`output`/`cached` in `history_store` (schema v3) so historical (non-today) `/analytics` days fold in delegated cost the same way live-scanned days do — previously the schema had no column for it, so the fold-in silently no-op'd for every historical day.

### Features
- **frontend:** Add a cost-by-day breakdown table to the Analytics page, mirroring the existing Agent breakdown table's style, sorted most-recent-first.

### DIFF
--- a/backend/history_store.py
+++ b/backend/history_store.py
@@ -36,7 +36,7 @@ from tt_paths import data_dir
 
 _log = logging.getLogger("tokentelemetry.history")
 
-SCHEMA_VERSION = 2
+SCHEMA_VERSION = 3
 
 # Sub-dicts folded into ``ecosystem_json`` and expanded back out on read. These
 # are the keys the analytics aggregation + delegation views consume beyond the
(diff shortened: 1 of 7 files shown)


backend/history_store.py
```diff
@@ -36,7 +36,7 @@
 
 _log = logging.getLogger("tokentelemetry.history")
 
-SCHEMA_VERSION = 2
+SCHEMA_VERSION = 3
```
AmirF194, line 39: Ran the migration path against a container in a clean Python 3.11 image.

`main` picked up its own unrelated `if ver < 3:` migration (project-path canonicalisation, `history_store.py:136`) while this branch was open, so both `main` and this PR now claim `SCHEMA_VERSION = 3` for two different things. Any existing install that already ran main's canonicalisation has `user_version=3`, so this PR's `if ver < 3:` block (the `ALTER TABLE ... ADD COLUMN delegated_cost` etc.) never runs against it:

```
$ python3 repro.py
after main._connect(): user_version = 3
PR #250 SCHEMA_VERSION constant: 3
history upsert failed: table sessions has no column named delegated_cost
upsert_sessions RAISED: OperationalError no such column: delegated_cost
```

`upsert_sessions` wraps the whole batch in one try/except (`history_store.py:198-289`, "store must never break the scan"), so the first row hitting this aborts the entire batch and the failure is only logged, never raised. On an install already at v3, every scan's session persistence silently no-ops, not just the delegated columns.

#338 already fixes this correctly by rebasing onto v4 instead of reusing v3 (`SCHEMA_VERSION = 4`, its own `if ver < 4:` block for the delegated columns) and looks like the one to carry forward. If you keep working from this branch instead, it needs the same version bump.

VasiHemanth:
Superseded by #338, merged as 9256f14.

#338 carries these same three commits rebased, plus fixes for the collisions that had built up since August: the delegated columns had to move to schema v4 because main took v3 for project canonicalisation, and the cache version to v14 because main took v13.

The fix itself is shipped and verified: on a real dataset the analytics total moved from $3,326.73 to $3,607.82 across 1,348 sessions, exactly the $280.99 of Claude subagent spend that was being shown beside the total and left out of it. The history-store half mattered as much as the aggregation half, since every day but today is served from SQLite and the fold-in was a silent no-op there.

Thanks for finding the undercount and for the independent verification in the commit messages. Credited in the release notes.

PR #245, opened 2026-08-06
feat(frontend): persist and restore scroll state across page navigations

hwantage:
## Summary
This PR addresses a key usability limitation where navigating back from session or project detail views to long list feeds (e.g., Dashboard, Projects list, Project Activity, and Hermes sessions) causes the scroll position to reset to the top.

### What was done
1. **Scroll State Persistence & Restoration**: Added `useScrollState` custom hook and `pageState` / `useSessionState` helper utilities that cache window scroll positions and inner container scroll elements (`sessionStorage`).
2. **Data-Ready Restoration**: Ensures scroll positions are restored smoothly only after page data fetching completes (`ready` state flag).
3. **Clean Reset on Primary Navigation**: Integrated `clearPageState()` into primary left navigation links (`Navigation.tsx`) so that clicking menu links cleanly resets cached scroll positions for a fresh browsing state.

---

### Development Journey & Trial/Error
* **URL-based History Back Attempt**: Initially, an approach using URL search parameters / query strings was consideredto handle history back behavior. However, sync issues arose when combining scroll parameters with active search queries, filter states, and keeping URLs clean for sharing.
* **Current Approach**: Transitioned to the `sessionStorage`-backed custom hook architecture. This cleanly decouples scroll state management from URL params.
* **User Experience**: In my local environment, this implementation provided a noticeably smoother and far more satisfying exploration experience when inspecting multiple sessions back-and-forth.

---

### Trade-offs & Concerns
1. **Maintenance Rules for New Features**: When adding new list views or stateful filter controls in the future, developers need to adhere to the following state management guidelines:
   * **Rule 1**: Assign a unique `pageKey` string for each page or scrollable container (e.g., `key_dashboard_recent_activity`).
   * **Rule 2**: Pass the `ready` boolean condition (e.g., `!loading && items.length > 0`) to `useScrollState` so restoration occurs after DOM rendering completes.
   * **Rule 3**: Call `clearPageState()` on top-level navigation triggers to ensure state resets when changing contexts.
2. **Browser Testing Scope**: Tested primarily on **Brave Browser** and **Safari** on macOS. Further cross-browser testing (Firefox, Windows Chrome/Edge) may be required.

## Type of change
- [ ] Bug fix
- [ ] New feature (new agent support, new dashboard view, etc.)
- [v] Improvement / refactor
- [ ] Docs / chore

## How to test
1. Launch TokenTelemetry (`./start.sh`)
[...]

PR #252, opened 2026-08-08
feat(hermes): persist session explorer filter state, simplify compone…

### LINKED ISSUE #251: [FEAT] Hermes Session Explorer: persist filter state via sessionStorage and fix scroll reset on history back
## Is your feature request related to a problem?
On the **Hermes → Session explorer** page (`/hermes/sessions`), the filter state(search, project, source, model, sort, page) is currently derived from URL query parameters. Unlike the Projects page (`/projects`), it does not persist its UI state between visits, and the back-navigation flow is inconsistent.

## Describe the solution you'd like
1. Session explorer filters should persist across navigations the same way the Projects page does (via `useSessionState`), so returning to the page keeps the user's filters and current page.
2. Scroll position should be retained and restored reliably on every history back navigation — including returning to `/hermes` — once the page data has finished loading.

## Which agent(s) does this relate to?
- [ ] Claude Code
- [ ] Gemini CLI
- [ ] Codex
- [ ] Cursor
- [ ] GitHub Copilot
- [ ] Qwen
- [ ] OpenCode
- [ ] Vibe
- [ ] Antigravity
- [x] All / General

## Additional context
I am happy to implement this and open a PR.

hwantage:
## Summary
This PR addresses state persistence and scroll restoration issues on the Hermes Session Explorer (/hermes/sessions) page, bringing its user experience in line with the Projects page while streamlining component architecture.

**Key Changes:**
1. Persist Filter State (useSessionState): 
- **Before**: The explorer derived filter states (search, project, source, model, sort, page) from URL query parameters on every render, causing filter and page selections to reset during navigation.
- **After**: Implemented useSessionState("hermes_sessions_page", DEFAULT_QUERY) so filters seamlessly persist in sessionStorage across navigations, matching the pattern used in /projects. Directly bound input elements to query.search and setQuery, removing redundant local state and timer boilerplate.

2. Fix Scroll Restoration & Race Condition (useScrollState):
- Added useScrollState to the Session Explorer to preserve scroll positions upon back-navigation.
- Fixed a race condition in useScrollState where a pre-restore programmatic scroll clamp of <main> to 0 (occurring while initial page data was fetching) would overwrite stored scroll offsets, causing /hermes to reset to the top when navigating back.

## Type of change
- [ ] Bug fix
- [x] Improvement / refactor
- [ ] Docs / chore

## How to test
1. Run `./start.sh`
2. Open /hermes, scroll down, click View all sessions.
3. Apply filters / navigate pages, open a session, then press Back — the explorer keeps its filters and scroll.
4. Press Back again — /hermes now restores its previous scroll position instead of jumping to the top.

## Checklist

## Related issue
Closes #251

### DIFF
--- a/frontend/src/components/hermes/HermesSessionExplorer.tsx
+++ b/frontend/src/components/hermes/HermesSessionExplorer.tsx
@@ -16,6 +16,7 @@ import {
 import { formatCost, formatTokens } from "@/lib/format";
 import { Badge, Button, Card, EmptyState, Skeleton } from "@/components/ui";
 import SourceBadge from "@/components/SourceBadge";
+import { useScrollState } from "@/lib/useScrollState";
 
 const PAGE_SIZE = 50;
 
@@ -60,6 +61,9 @@ export default function HermesSessionExplorer() {
     pollMs: 15_000,
     initial: { sessions: [], pagination: { page: 1, page_size: PAGE_SIZE, total: 0, total_pages: 0 } },
   });
+
+  // Restore scroll position once session data is loaded.
+  useScrollState("key_hermes_sessions_page", !loading);
 
   const activeFilterCount = [query.project, query.source, query.model, query.search].filter(Boolean).length;
 
--- a/frontend/src/lib/useScrollState.ts
+++ b/frontend/src/lib/useScrollState.ts
@@ -41,6 +41,8 @@ export function useScrollState(key: string, isReady: boolean = true) {
 
   const saveScrollPosition = useCallback((top: number) => {
     if (isRestoringRef.current) return;
+    // Ignore programmatic scrolls that fire before the initial restore runs
+    if (!restoredRef.current) return;
     sessionStorage.setItem(`scroll_${key}`, String(top));
   },[key]);
 


VasiHemanth:
Thanks for this, and for the detailed writeup. The scroll-restoration gap on `/hermes/sessions` is real, and you've correctly identified a filter problem that does exist. Some context that isn't visible from the outside, plus what I found testing the branch.

**On moving filters from the URL to sessionStorage.** This was a deliberate choice rather than an oversight. `useSessionState` and `useScrollState` landed in `lib/` in #247 (commit `4baddbe`, your #245 work), and I wrote `HermesSessionExplorer.tsx` about half an hour later in that same PR — with the sessionStorage helpers already sitting in the tree — using `useSearchParams` + `router.replace` and a `returnPath` carrying the query string. I wanted explorer views to be copy-pasteable and bookmarkable. Same for `profileScope` on `/hermes`: when I cherry-picked that file from #245 I kept `useState` on purpose. This PR reverses both, which is why I want to settle the design question before the code.

**The persistence doesn't hold up in the journey it's meant to fix.** I ran the branch against its merge-base with the same backend. Every sidebar link calls `clearPageState()` (`Navigation.tsx:131`), which purges every key registered through `registerStateKey` — including `hermes_sessions_page`. So: filter the explorer, click any sidebar item, come back through "View all sessions", and the filters are gone, because the state was deleted on the way out. I watched `hermes_sessions_page` go from `search: "marketing"` to `null` on a single sidebar click.

What survives is the case where you never touch the sidebar — the in-page back arrow, browser Back, a reload. Those are the paths the URL already handled, since the filters were in the address bar. So the net gain over main is small, while the address bar stops being shareable.

**Measured regression.** Typing a 9-character search fires 9 requests to `/hermes/sessions` on this branch versus 1 on main — the 300ms debounce went out with the "timer boilerplate". Worth noting that `project`/`source`/`model` were already un-debounced on main, so the right fix isn't reverting: it's debouncing all four inputs, which would leave the page better than it is today.

**Two things that need fixing regardless of the design call.** Deleting `readQuery` also removed the only input sanitizer — the `Math.max(1, …)` clamp on `page` and the whitelist on `sort`. Since `useSessionState` merges `{...defaults, ...stored}`, a stale entry wins over the defaults, so a bad `page` or a retired `sort` reaches the backend and 422s. The escape hatch then doesn't render: `activeFilterCount` counts only project/source/model/search, so the Clear button is hidden exactly when a bad `page` is the problem.

**On the scroll fix.** I couldn't reproduce the reported reset — a full `/hermes` → explorer → session detail → Back → Back cycle restores the scroll offset correctly on main here (22 sessions, warm cache). The mechanism you describe is plausible under a slower initial load, and the guard itself is harmless and correctly placed, so I'm not disputing it; I just can't confirm it.

**Where this leaves us.** I've opened #258, which takes a different route to the filter problem you found: the URL stays the source of truth, and the last filter set is remembered separately and applied only when the explorer opens with no query of its own. A shared link still wins, and a sidebar round-trip keeps your filters. It also replaces the pagination with a "Load more" list, which removes the `page` value behind the sanitizer issue above.

That leaves the half of #251 I haven't done: **scroll restoration on `/hermes/sessions`**. None of the six existing `useScrollState` call sites cover the explorer, so that's a real gap and I'd like it. If you're up for it, a focused PR with just that — no `profileScope` change, no sessionStorage filter swap — would land quickly. The `restoredRef` guard is welcome alongside it if you can attach a repro, since it edits the shared hook that all seven call sites depend on and I couldn't trigger the reset here.

Your last three PRs went in clean and this one is well-built too. The friction is architectural direction, not code quality, and the underlying problem you reported was real — thanks for chasing it.

hwantage:
Thanks for the detailed feedback and for opening #258!

I completely agree with keeping the URL as the primary source of truth for shareability and avoiding the sidebar filter-clearing side effect.

I have refactored this PR to focus solely on **scroll restoration**:
1. Reverted all filter-related changes so they can be handled via #258.
2. Added `useScrollState` to `/hermes/sessions` so returning from session details restores the scroll position.
3. Kept the `restoredRef` guard in `useScrollState.ts` to prevent scroll loss when navigating between `/hermes` and `/hermes/sessions`.

---

### Exact Bug Scenario & Technical Explanation of `restoredRef` Guard

#### Reproduction Steps (The Failing User Flow)
1. **Scroll on Dashboard (`/hermes`)**:
   - Go to the main Hermes dashboard (`/hermes`).
   - Scroll down to the bottom section (e.g., *Recent Hermes sessions* table or *Telemetry* section).
   - `sessionStorage` saves the current scroll position for `key_hermes_page` (e.g., `1200px`).
2. **Navigate to Sessions (`/hermes/sessions`)**:
   - Click the **"View all sessions"** link to navigate to `/hermes/sessions`.
   - Scroll down through the session explorer list.
3. **Navigate Back (Browser 'Back' Button)**:
   - Click the browser **Back** button to return to `/hermes`.
4. **The Race Condition (Why `/hermes` scroll was lost)**:
   - Upon navigating back to `/hermes`, the component mounts and initiates async resource fetching (`sessionsRes`, `overviewRes`). During this time, `loading = true` (`isReady = false`).
   - Before data arrives, the browser mounts the empty/skeleton layout and resets the container scroll top to `0`, firing an unpreventable native `scroll` event.
   - **Without `restoredRef` guard**: The hook's `scroll` listener catches this `scrollTop = 0` event while `isReady` is still `false`, and writes `0` into `sessionStorage` under `scroll_key_hermes_page`.
   - **Result**: When data loading completes (`loading = false` -> `isReady = true`), the saved position in `sessionStorage` has already been overwritten with `0`. The `/hermes` dashboard fails to restore to `1200px` and snaps back to the top of the page.

---

#### How `restoredRef.current` Fixes This Issue

```typescript
const saveScrollPosition = useCallback((top: number) => {
  if (isRestoringRef.current) return;
  // Ignore programmatic/layout scrolls that fire BEFORE the initial restore runs
  if (!restoredRef.current) return;
  sessionStorage.setItem(`scroll_${key}`, String(top));
}, [key]);
```

1. **Guarding Pending Restoration (`restoredRef.current = false`)**:
   - While returning to `/hermes` and waiting for initial data (`isReady = false`), all scroll events triggered by initial DOM mounting/layout resets are ignored.
   - The stored scroll offset (`1200px`) in `sessionStorage` remains untouched.
2. **Restoring Intact Scroll Position (`isReady = true`)**:
   - Once data loading finishes, the hook reads the preserved `1200px` from `sessionStorage`, scrolls `/hermes` back down to `1200px`, and sets `restoredRef.current = true`.
3. **Resuming Normal Save**:
   - After `restoredRef.current` becomes `true`, subsequent user-initiated scrolls update `sessionStorage` as expected.

Ready for review!

VasiHemanth:
Closing in favour of #313, which supersedes this one.

#313 carries this PR's changes verbatim: the same guard in `useScrollState.ts` (`if (!restoredRef.current) return` in `saveScrollPosition`) and the same `useScrollState("key_hermes_sessions_page", !loading)` call in the explorer. It then adds the piece that makes the restore hold past the first screenful.

That extra piece turns out to be necessary rather than optional. On `main`, the row count lives in local React state (`const [rowLimit, setRowLimit] = useState(PAGE_SIZE)`), so leaving the page and coming back remounts the component and resets it to 50. Restoring a scroll offset taken at 150 rows then lands on a list that is only 50 rows tall, and there is nothing to scroll to. Moving `limit` into the URL is what gives the restore a list to land in.

#313 is also rebased on the current explorer, which picked up the remembered-filters change after this PR was opened.

So this is closed as included, not rejected. The scroll-restoration fix ships as you wrote it. Thanks for both passes at it, and for splitting the follow-up out cleanly.

PR #285, opened 2026-08-21
fix(grok): read billed usage from the unified inference log

sirenexcelsior:
## Summary
- Grok session directories only store a context-window footprint (`contextTokensUsed`), so the dashboard treated that number as Input and left Output, Cached, and cost empty or near zero.
- The Grok CLI already writes per-request `prompt_tokens` / `cached_prompt_tokens` / `completion_tokens` on `shell.turn.inference_done` rows in `~/.grok/logs/unified.jsonl`. The scanner now aggregates those by session id.
- Pricing adds `grok-4.6` at the published list rates and applies xAI's per-request 200k-prompt 2× cliff. Fuzzy lookup no longer maps `grok-4.6` onto `grok-4`.
- The session header shows Input / Output / Cached when billed usage is present (`tokens.source === "usage"`). Sessions with no log rows still show Context and em-dashes.

## Type of change
- [x] Bug fix
- [ ] Improvement / refactor
- [ ] Docs / chore

## How to test
1. Run `./start.sh` on a machine that has used Grok Build.
2. Open a Grok session that appears in `~/.grok/logs/unified.jsonl`.
3. Confirm the header shows Input, Output, and Cached (not Context / —) and a non-trivial API-equivalent cost.
4. Open a Grok session with no matching log rows — it should still show Context and em-dashes.
5. `pytest backend/test_grok_usage.py`

## Checklist

## Related issue
N/A

Made with [Cursor](https://cursor.com)

### DIFF
--- a/backend/main.py
+++ b/backend/main.py
@@ -435,6 +435,9 @@ HERMES_PROFILES_DIR = HERMES_DIR / "profiles"
 # Stores rich per-session data under ~/.grok/sessions/<encoded-cwd>/<uuid>/
 GROK_DIR = HOME / ".grok"
 GROK_SESSIONS_DIR = GROK_DIR / "sessions"
+# Per-turn billed usage (prompt / cached / completion) lives here, not in the
+# session dir. Session files only expose a context-window footprint.
+GROK_UNIFIED_LOG = GROK_DIR / "logs" / "unified.jsonl"
 
 # Grok Build session file names (per <cwd-uuid> directory)
 GROK_SUMMARY = "summary.json"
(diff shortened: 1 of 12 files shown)


VasiHemanth:
Validated this against real Grok data rather than by inspection: 285 session
dirs under `~/.grok/sessions/` and a live `~/.grok/logs/unified.jsonl`
(2.68 MB, 1,272 `shell.turn.inference_done` rows, 2026-08-05 to 2026-08-21).
It holds up. Notes below for the record.

**The join is exact.** 96 distinct `sid` values in the log, 96/96 match a
session directory name, zero orphans. The remaining 189 of 285 dirs predate
the log window and take the context-footprint fallback, as intended.

**Cost recomputed independently, matches to ten decimal places.** I re-derived
it straight from the raw JSONL using xAI's published two-tier table, bypassing
`pricing.py` entirely, with both paths pinned to one frozen snapshot so the
live-appending log couldn't move underneath:

| 96 usage-path sessions, 1,272 turns | this PR | independent recompute |
|---|---|---|
| input | 36,794,003 | 36,794,003 |
| output | 825,971 | 825,971 |
| cached | 164,609,280 | 164,609,280 |
| cost | $185.8286 | $185.8286 |

Cost delta $0.0000000000.

What it changes across the whole corpus: output 0 → 815,760, cached
0 → 162,397,696, total cost $65.03 → $215.24. Output and cache-read were
invisible for every Grok session before this.

**Rates check out against the primary source.** grok-4.6 $2.00/$0.50/$6.00,
grok-4.5 $2.00/$0.30/$6.00, grok-4.3 $1.25/$0.20/$2.50, and the ≥200k tier is
exactly 2x on all three rates, so the single `* 2.0` in
`calculate_xai_turn_cost` is the right shape rather than a coincidence. 28.6%
of local turns cross the threshold, so it is not a corner case.

**The fuzzy-match guard has no blast radius.** `_fuzzy_key_matches` changes
`calculate_cost` for every provider, so I diffed main vs this branch over the
full effective table plus generated variants: 15,566 model ids. 32 change
cost. All 32 are grok. The headline one is `grok-4.6`, which was matching
`grok-4` and billing at $3/$15.

**Also verified**

- `reasoning_tokens < completion_tokens` in all 1,272 rows, so reasoning is a
  subset of completion and correctly stays out of `output` and out of cost.
- No subagent double-count: 94 of the 96 usage-path sessions are subagent
  children, each with its own `sid`, so parent and child token sets are
  disjoint and `tokens_recorded: False` still holds.
- The new `tokens["source"]` string can't reach any `sum(tokens.values())`
  site in `main.py`; all of them operate on other agents' local dicts.
- Log parse is 4.3 ms cold, 0.02 ms warm. The mtime+size cache key works.
- The nine one-line test changes plug a real leak: without `GROK_UNIFIED_LOG`
  in those hermetic `scan_env` fixtures, the tests read the developer's actual
  Grok log.
- Full backend suite: 452 passed on this branch vs 446 on main, with a
  byte-identical set of 22 pre-existing failures (unrelated local-config
  pollution in `test_power_config` / `test_pricing` / `test_update_check`).
  `tsc --noEmit` fails identically on both branches on a missing local
  `@lobehub/icons` install.

**Scope of the check:** every session on this machine records
`current_model_id: grok-4.5`, so the empirical cost comparison covers that one
model. The rates and the cliff were confirmed from the docs for 4.6 and 4.3 as
well.

**Two small things, neither blocking, both pre-existing rather than introduced
here.** I have them on a follow-up branch so they don't hold this up:

1. `pricing.py`'s inline `grok-build` / `grok-build-0.1` entries carry
   grok-code-fast-1's rates ($0.20 in / $1.50 out). xAI lists grok-build-0.1
   at $1.00 in / $0.20 cached / $2.00 out, and the bundled table already had
   it right as `xai/grok-build-0.1`, so the inline entry was shadowing the
   correct figure.
2. `_scan_grok_sessions` reassigns `model` from `signals.modelsUsed` after the
   token block has already priced the session. A session with no
   `current_model_id` but a populated `modelsUsed` displays one model and is
   costed as `grok-build`. Unreachable on this data (`modelsUsed` is empty in
   all 285 local sessions).

One observation, no change proposed: `_grok_price_turns` applies the 200k
cliff to any grok model. That is right for everything xAI currently lists, and
defaulting it on is the safer failure mode, since a newly released model that
missed an opt-in list would silently undercount. Retired ids like
`grok-code-fast-1` would be doubled with nothing behind them, but they no
longer appear on the pricing page and none exist locally.

Nice find on the `sid` key. Thanks for digging this one out.

PR #286, opened 2026-08-21
fix(hooks): judge the branch the push actually comes from

VasiHemanth:
## Summary

Both pre-push hooks resolved the repo with `git rev-parse --show-toplevel`
from their own process cwd, which is the session's project root. That is not
where the push runs once Claude is in a worktree, or once the command starts
with `cd <dir> && ...`, so a worktree push was judged against whatever branch
the project root happened to be sitting on.

This repo does most of its work in worktrees, so the hooks were routinely
answering about the wrong branch.

## What went wrong in practice

A worktree branch carrying only `fix:` commits was denied, because the project
root sat on a `feat:` branch whose `UPDATE.json` change was still uncommitted.
Per `.claude/CLAUDE.md` case 1, a pure `fix:` branch should have been allowed
silently.

The same path also denied a worktree checked out on `main`, despite the hook's
explicit "skip enforcement if we're on main" rule, because the branch it read
was never `main` to begin with.

## The fix

Adds `resolve_push_cwd()` to `enforce-update-json.py`; both hooks call it
before asking git anything. Precedence, most explicit first:

1. `git -C <dir>` in the push fragment itself
2. the accumulated `cd <dir>` prefix of the same command chain
3. the payload's `cwd` field, which is the worktree root after Claude enters a
   worktree
4. the hook's own cwd, i.e. today's behaviour

Relative paths compose against whatever is current at that point. Forms that
can't be resolved to a concrete path (`cd -`, `cd ~/x`) fall back rather than
guess.

Non-worktree pushes are unaffected: the payload cwd equals the process cwd, so
every path resolves exactly as it does today.

## Type of change

- [x] Bug fix
- [ ] Improvement / refactor
- [ ] Docs / chore

## How to test

`backend/test_hooks.py` (new, 15 tests) covers the resolver directly and
drives `enforce-update-json.py` end to end as a subprocess with a real
PreToolUse payload, over throwaway repos and real `git worktree` checkouts.

    pytest backend/test_hooks.py

Eleven of the fifteen fail against the unfixed hooks, including the two
behavioural ones (`test_worktree_push_is_judged_on_its_own_branch`,
`test_main_branch_is_always_allowed`). The guard tests pass either way, so the
fix doesn't loosen the gate:

- `test_denies_feat_branch_without_update_json`
- `test_worktree_feat_branch_still_denied` (a real `feat:` worktree is still denied)
- `test_allows_when_update_json_is_touched`
- `test_push_detection_unchanged`

Verified against this repo as well: the previously-denied push from a
`fix:`-only worktree is now allowed, and the live `feat/local-model-insights`
branch is still correctly denied.

Full backend suite: 461 passed, with the same pre-existing failure set as
`main` (22, all unrelated local-config pollution in `test_power_config` /
`test_pricing` / `test_update_check`).

## Checklist

- [x] I tested this locally

## Related issue

N/A. Found while pushing a follow-up to #285 from a worktree.

### DIFF
--- a/.claude/hooks/enforce-update-json.py
+++ b/.claude/hooks/enforce-update-json.py
@@ -165,6 +165,101 @@ def _split_on_chain_operators(s: str) -> List[str]:
     return fragments
 
 
+def _cd_target(tokens: List[str]) -> Optional[str]:
+    """If a command fragment is a bare `cd <dir>`, return <dir>, else None.
+
+    `cd` with no argument, `cd -`, and `cd ~...` are treated as unknown so the
+    caller falls back rather than guessing at a directory.
+    """
+    i = 0
+    while i < len(tokens) and re.match(r"^[A-Za-z_][A-Za-z0-9_]*=", tokens[i]):
+        i += 1
+    if i >= len(tokens) or tokens[i] != "cd":
+        return None
+    rest = [t for t in tokens[i + 1:] if not t.startswith("-")]
+    if len(rest) != 1:
+        return None
+    target = rest[0]
+    if target.startswith("~"):
+        return None
+    return target
+
+
+def _git_dash_c_target(tokens: List[str]) -> Optional[str]:
+    """Return the argument of `git -C <dir>` / `--work-tree=<dir>` if present."""
+    i = 0
+    while i < len(tokens) and re.match(r"^[A-Za-z_][A-Za-z0-9_]*=", tokens[i]):
+        i += 1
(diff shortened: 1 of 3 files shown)


VasiHemanth:
Superseded by #306, which includes the safer resolver and restored end-to-end regression coverage.

VasiHemanth:
Closing as superseded by #306. The replacement keeps the worktree-aware fix, adds safer directory validation, and now carries this PR’s stronger end-to-end hook coverage. Please review and merge #306 only.

PR #290, opened 2026-08-22
fix(backend): fold / vs \ separator variants into one project identity

Rub3nCT:
Agent CLIs log cwd in their own separator style: some emit C:\a\b, others C:/a/b, a few mix both. Projects were grouped by exact string, so one real Windows folder could surface as several project cards.

Introduce tt_paths.canonical_project() and apply it at every identity boundary:

- end of _scan_sessions_sync(): every session carries one identity, so rollups, filters and durable history agree
- canonical_repo()/_repo_worktree_paths(): emit canonical form too; otherwise worktree grouping synthesises duplicate repo cards on Windows (native-separator str(Path) vs forward-slashed cards)
- get_projects(): hidden-set membership compares canonically, so entries saved before the change keep working in either style
- harness_config hide/unhide: store the canonical form; unhide rebuilds the set so a legacy entry saved in another style is removed too instead of sticking forever
- budgets: per-project filters match across separator styles
- history_store v3 migration: one-time re-canonicalisation of stored rows (DISTINCT projects, idempotent) so analytics over pruned sessions keeps matching after upgrade; query() also accepts either style as filter input

Windows-shaped paths (drive-letter or UNC prefix) are unified to forward slashes with trailing separators trimmed. POSIX paths are only trailing-trimmed: a backslash is a legal filename character there. Letter case is intentionally not folded (case-sensitive filesystems). Sentinels ("unknown", "Antigravity / unassigned") pass through.

New backend/test_project_paths.py pins the behaviour end-to-end: two Claude transcripts logging C:\Users\dev\proj and C:/Users/dev/proj now yield exactly one card holding both sessions. It runs under pytest and standalone (python backend/test_project_paths.py), no new deps. Two smallcode assertions move to canonical comparison for the same reason.

### DIFF
--- a/backend/harness_config.py
+++ b/backend/harness_config.py
@@ -21,7 +21,7 @@ import uuid
 from pathlib import Path
 from typing import Any, Dict, List, Set
 
-from tt_paths import data_dir
+from tt_paths import canonical_project, data_dir
 
 # Resolved once at import. The data dir is an install-time setting (exported via
 # TOKENTELEMETRY_DATA_DIR / TOKENTELEMETRY_HOME before the backend launches), so
@@ -118,16 +118,21 @@ def save_hidden(paths: Set[str]) -> None:
 
 def hide_project(path: str) -> Set[str]:
     current = load_hidden()
-    current.add(path)
+    # Store the canonical form so membership checks against canonicalised
+    # card paths match regardless of the separator style that was submitted.
+    current.add(canonical_project(path))
     save_hidden(current)
     return current
 
 
 def unhide_project(path: str) -> Set[str]:
-    current = load_hidden()
-    current.discard(path)
-    save_hidden(current)
-    return current
+    # Rebuild the set in canonical form while removing the target, so an
+    # entry saved under another separator style (pre-canonicalisation) is
+    # removed too instead of surviving as an unremovable stale entry.
+    target = canonical_project(path)
+    updated = {p for p in (canonical_project(c) for c in load_hidden()) if p != target}
+    save_hidden(updated)
+    return updated
 
 
 def load_preferences() -> Dict[str, Any]:
--- a/backend/history_store.py
+++ b/backend/history_store.py
@@ -32,11 +32,11 @@ from datetime import datetime, timezone
 from pathlib import Path
 from typing import Any, Dict, Iterable, List, Optional, Sequence, Set, Tuple
 
-from tt_paths import data_dir
+from tt_paths import canonical_project, data_dir
 
 _log = logging.getLogger("tokentelemetry.history")
 
-SCHEMA_VERSION = 2
+SCHEMA_VERSION = 3
 
 # Sub-dicts folded into ``ecosystem_json`` and expanded back out on read. These
 # are the keys the analytics aggregation + delegation views consume beyond the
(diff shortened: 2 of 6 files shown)


VasiHemanth:
Validated this against `main` before reviewing. The bug is real and this fixes it.

Reproduced on `main` with two Claude transcripts logging the same Windows folder in opposite separator styles: two `/projects` cards, one session each. On this branch: one card, two sessions. I also exercised the schema v3 migration on a real v2 database built by `main`'s own code, since that is the part that touches persisted data: 3/3 rows preserved, token totals preserved, idempotent on re-open, and old-style filters still match. Full backend suite shows no regressions (identical failure set to `main`, plus your 11 new tests passing).

Putting the fold at the end of `_scan_sessions_sync` rather than at each of the sites that set `project` is the right call. The per-session `scan_cache` sidecars on disk survive restarts and still hold pre-canonicalisation strings, and because they are consumed well before that loop, they get corrected on the way out. A per-source fix would have left them producing duplicates indefinitely.

One gap worth raising before this is read as covering more than it does.

### Cross-agent duplicates on Windows are still split

`_WIN_PATH_RE` is anchored `^(?:[A-Za-z]:|\\\\)`. The Copilot/VS Code and Cursor scanners do not produce that shape. They build the project path from the workspace `file:` URI:

```python
project_path = unquote(folder_url.replace("file://", ""))
```

On Windows VS Code writes `file:///c%3A/Users/dev/proj`, which decodes to `/c:/Users/dev/proj`: leading slash, lowercase drive. The regex never matches, so the path passes through untouched.

Seeding one Claude session and one Copilot session in the same Windows folder, on this branch:

```
agent=claude   project='C:/Users/dev/proj'
agent=copilot  project='/c:/Users/dev/proj'

/projects CARDS: 2
```

This is probably the duplicate a Windows user is most likely to hit, since it needs only two different agents rather than one agent logging inconsistently.

It looks like two lines inside the branch you already have: strip a leading `/` when it precedes a drive letter, then upcase the drive letter.

On the drive-letter case specifically, I think the docstring's reasoning does not apply. It justifies not folding case by pointing at case-sensitive filesystems, but control only reaches that branch after `^[A-Za-z]:` has matched, so the path is already established as Windows, where the drive letter is case-insensitive by definition. Folding just the drive letter inside that branch looks safe; the rest of the path is what the argument is actually protecting and can stay as it is.

Either fixing it here or noting in the description that the `file:` URI form is out of scope works. I mainly want it not to be mistaken for already handled.

### Smaller notes, non-blocking

- `canonical_project("C:/")` returns `"C:"`. On Windows that means "current directory on drive C", not the drive root, so it is corruption rather than a missed fold. The guard covers the POSIX root (`/`, `//`) but not the drive root.
- `_session_in_filters` still compares `projects` verbatim while `history_store.query` now canonicalises them, so the two halves of a merged `/analytics` view apply different filters. Unreachable from the UI today (the analytics page never sends `projects`), so this is symmetry rather than a bug.
- Project cards render `path` verbatim, so a Windows card that reads `C:\Users\dev\proj` today will read `C:/Users/dev/proj` after this lands. Fine as a trade, but worth a line in the description since a user will notice.

Nothing above blocks merging. Thanks for the thorough tests and for explaining in the comments what you deliberately chose not to fold.

PR #299, opened 2026-08-26
fix: harden local launch and clear frontend lint baseline

jjoanna2-debug:
## Summary

- bind the Next.js dev server to the configured host so the default launch is actually loopback-only
- remove the shell-based npm spawn and align the documented Node requirement with Next.js 16
- clear the frontend ESLint baseline (168 errors and 24 warnings to zero) while retaining narrow, documented exceptions for polymorphic trace payloads and hydration sync
- update the vulnerable `brace-expansion` and `js-yaml` transitive tooling packages
- pin the Turbopack workspace root to the frontend package

The changes are split into three commits so launcher hardening, dependency updates, and lint cleanup can be reviewed independently.

## Validation

- `npm run lint` — 0 errors, 0 warnings
- `npx tsc --noEmit`
- `npm run build`
- `npm audit --audit-level=high` — 0 vulnerabilities
- live dashboard and API returned HTTP 200 on alternate ports, with both listeners verified on `127.0.0.1`
- two read-only regression-review passes completed; final pass reported no findings

### DIFF
--- a/README.md
+++ b/README.md
@@ -192,7 +192,7 @@ Captured plan-mode outputs from Claude Code's `/plan` command and equivalent in 
 
 ## Requirements
 
-- **Node.js 18+**
+- **Node.js 20.9+**
 - **Python 3.9+**
 - **git**
 - Any supported AI coding agent already installed (Claude Code, Gemini CLI, Codex, etc.)
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -223,8 +223,10 @@ function which(cmd) {
 }
 
 function checkNode() {
-  const major = parseInt(process.versions.node.split('.')[0], 10);
-  if (major < 18) die(`Node.js 18+ required (detected ${process.versions.node}).`);
+  const [major, minor] = process.versions.node.split('.').map(Number);
+  if (major < 20 || (major === 20 && minor < 9)) {
+    die(`Node.js 20.9+ required (detected ${process.versions.node}).`);
+  }
 }
 
 function findPython() {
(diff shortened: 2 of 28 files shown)


VasiHemanth:
Merged. Recording the independent verification here since the loopback finding is the interesting part and it deserves more than "LGTM".

## The exposure was real, and reproduced

Before the fix, on a default launch, the two services disagreed about what "local" meant:

```
node    *:3000            frontend, every interface
Python  127.0.0.1:8000    backend, loopback only
```

Fetching the dashboard over this machine's own LAN address returned **HTTP 200**. After the fix, same request, **connection refused**, while `127.0.0.1:3000` kept returning 200. So this was not a cosmetic change to what `lsof` prints. The dashboard really was being served to the local network on a default launch, and the token gate did not cover it because the gate protects the API, not the Next dev server.

## Remote mode still works

The obvious risk with this fix is nailing the frontend shut and breaking tailnet access. Checked with `--host 0.0.0.0`:

```
Python  *:8200
node    *:3200
```

Reachable over the LAN address, HTTP 200. Both processes now follow the host you asked for, and neither of them opts you in on its own.

## Node 20.9

Correct, and it is a correction rather than a preference. `next/package.json` on the pinned 16.2.9 declares `engines: { node: ">=20.9.0" }`, so the previous "Node.js 18+" in the README and `checkNode()` was promising something that could not work.

## One follow-up, already handled

The dependency bump in commit 2 would not have reached anyone who already had `node_modules`. `ensureFrontend()` stamped `sha1(package.json)` and skipped `npm ci` on a match, but this PR moves `frontend/package-lock.json` without touching `frontend/package.json`, so the stamp matched and the install was skipped. Measured on a tree carrying the old stamp and this PR's lock:

```
sha1(package.json) = 13848cc7bc9f...
stamp              = 13848cc7bc9f...
MATCH -> npm ci skipped

brace-expansion 1.1.16   (this PR intends 1.1.18)
js-yaml         4.3.0    (this PR intends 4.3.1)
```

Not a problem with this PR, a pre-existing gap it happened to expose: `npm ci` installs what the lock pins, so the lock was the input that decided the outcome and `package.json` was the file being watched. #300 fixes the stamp to hash both, and landing it before this PR means the bump reaches existing installs on their next launch rather than only fresh ones. Confirmed end to end after that: `npm ci` ran, and reported `brace-expansion 1.1.18`, `js-yaml 4.3.1`, `found 0 vulnerabilities`.

Thanks for the loopback catch in particular. It had been sitting in the default launch path, invisible because nothing was broken from the user's seat.

PR #306, opened 2026-08-29
fix(hooks): judge the branch being pushed, not the session directory

VasiHemanth:
## Summary

Both local pre-push hooks previously resolved Git state from the Claude session directory. A command such as cd worktree and then git push could therefore validate a different branch than the one being pushed.

This PR resolves the actual push directory by following command-chain cd prefixes and git -C, validates the resolved directory, and shares that logic with the local security reviewer.

It prevents both false denies on clean fix worktrees and false allows for feature worktrees missing UPDATE.json.

## Verification

The self-contained regression suite now runs 15 checks, including real PreToolUse hook subprocess tests for:

- feature branches with and without UPDATE.json
- fix-only worktrees through cd, git -C, and payload cwd
- feature worktrees without UPDATE.json
- main worktrees
- malformed paths, quoted commands, and push detection

Run: python3 backend/test_hook_push_cwd.py

## Supersedes

Supersedes #286, with its end-to-end regression coverage restored here.

### DIFF
--- a/.claude/CLAUDE.md
+++ b/.claude/CLAUDE.md
@@ -62,6 +62,9 @@ This is enforced by `.claude/hooks/enforce-update-json.py`, wired up via
 
 1. Detects `git push` / `gh pr create` invocations (via proper shell
    tokenisation — not substring matching, so `echo "git push"` is fine).
+   Then resolves **which directory the push runs in** — following a
+   `cd <dir> &&` prefix or `git -C <dir>` — because the hook process's own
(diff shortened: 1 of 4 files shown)


VasiHemanth:
This is now the single merge candidate for the worktree-aware hook fix. It supersedes #286 and includes its stronger end-to-end regression coverage: 15 self-contained checks exercise the real PreToolUse hook for feature, fix-only worktree, and main-worktree cases. Verified locally with `python3 backend/test_hook_push_cwd.py`.

PR #329, opened 2026-09-07
feat: add Kimi Code agent support

kthreatt:
Adds [Kimi Code](https://www.kimi.com/) (Moonshot AI's coding CLI) as a fully supported agent — the 18th coding agent tracked.

## What it does
- **Session scanning**: reads `~/.kimi/sessions/<workdir-hash>/<session-uuid>/wire.jsonl`; token usage comes from `StatusUpdate` records (`input_other` / `output` / `input_cache_read` / `input_cache_creation`), de-duped by `message_id`. Projects are resolved via `~/.kimi/kimi.json` `work_dirs`; titles from `state.json` `custom_title` with first-user-prompt fallback; model from `config.toml` `default_model` (falls back to `kimi-for-coding`). Cost uses the existing Moonshot pricing table.
- **Panel**: `build_kimi` surfaces `loop_control` config and the work-dir registry. Never reads `~/.kimi/credentials/`.
- **Summarizer**: new `summarizers/kimi.py` adapter shells `kimi -p` headlessly.
- **Billing**: Kimi subscription plan in `AGENT_PLANS` (membership bucket drains first, API paygo fallback).
- **Frontend/menubar**: agent key, brand colors (theme-aware monochrome), "Ki" monogram, docs in `supported-agents.mdx`, README/llms.txt lists, UPDATE.json entry.

## Testing
- New `backend/test_kimi_scan.py` (11 tests: token sums, de-dupe, titles, model fallback, project mapping, cost, empty/malformed dirs, detection, trace endpoint) plus updated guardrail tests. Targeted files: 200 passed. Full backend suite: 785 passed; remaining failures are pre-existing environment-dependent cases, byte-identical on clean HEAD.
- Frontend: eslint clean on touched files, `tsc --noEmit` clean, lib tests 29/29.
- Also verified against a real `~/.kimi` install (live sessions scanned correctly; summarizer invoked successfully).

### DIFF
--- a/.claude/skills/change-brief/SKILL.md
+++ b/.claude/skills/change-brief/SKILL.md
@@ -1,0 +1,270 @@
+---
+name: change-brief
+description: Explain a change that already exists — a pull request, a merged commit, a closed issue's fix, or an uncommitted diff — in terms of the FEATURE it touches and how someone actually uses that feature, then the bug or gap it addresses. Always states who wrote it (Claude in an earlier session, an outside contributor, or the maintainer), right after the opening model and before any mechanism. Use when the user pastes a PR link or number, says explain this PR / what does this PR do / what is 331 / what's this commit / explain these changes / what did you change, asks "in simple words" or "explain the scenario", or wants to understand work they did not write themselves. Always opens with a concrete everyday analogy and a one-sentence plain statement of the defect; plain language is the DEFAULT output, never a mode the user has to ask for. Read-only: it reads the diff, the surrounding code and git history, and never runs the app, never checks out a branch, never edits. NOT for deciding whether to build something not yet written (use /issue-brief) and NOT for finding defects (use /code-review).
+---
+
+# change-brief — explain a change that already exists
+
+`/issue-brief` explains something **not yet built**, to decide whether to build
+it. This skill explains something **already written**, so the reader
+understands what it does and whether to merge or trust it.
+
+The reader did not write this code. Often nobody they can ask wrote it — it may
+be an outside contributor's PR, or something Claude produced in a session they
+have since forgotten. Assume no memory of the change and no file layout in
+their head.
+
+## 1. Resolve the target
+
+| They said | Fetch with |
+|---|---|
+| PR number or URL | `gh pr view <n> --json number,title,author,body,files,additions,deletions,mergeable,mergeStateStatus,commits` |
+| "what is 331" (bare number) | Try `gh pr view <n>` first; fall back to `gh issue view <n>` |
+| commit SHA | `git show --stat <sha>` |
+| "these changes" / "what did you change" | `git status --short` and `git diff` (plus `git diff --staged`) |
+| merged issue | `gh issue view <n>`, then find the PR that closed it |
+
+Get the **real** diff, not the description. A PR opened weeks ago is measured
+against today's base:
+
+```bash
+git fetch origin main pull/<n>/head:pr<n>
+git diff origin/main...pr<n> --stat
+```
+
+A `git diff main...` against a stale local `main` reports hundreds of unrelated
+files. If the stat looks absurdly large for the PR's description, that is the
+cause — fetch `origin/main` and re-diff before believing it.
+
+## 2. Establish authorship — this is not optional
+
+The user asked for this explicitly: **never explain a change without saying who
+wrote it.** They need to know whether they are reading their own past decision,
+a contributor's proposal, or Claude's output from a session they don't recall.
+
+It lands in position 2, immediately after the opening model (§3 step 1) and
+before any mechanism. That is a deliberate change: it used to go first, and a
+brief that opened with a username, three PR numbers and a timestamp got
+"explain it clearly, I still didn't understand" in reply. A name means nothing
+until the reader knows what the thing does. Two lines, still never skipped.
+
+- **The change itself:** `gh pr view <n> --json author`, or
+  `git log -1 --format='%an <%ae>' <sha>`.
+- **Claude's fingerprint:** commits Claude wrote carry a
+  `Co-Authored-By: Claude` trailer and a `Generated with Claude Code` line.
+  `git log --format='%H %an %s' -- <file> | head` shows who has been touching
+  the area.
+- **The underlying feature is usually older than the change.** A PR fixing a
+  bug in a feature did not create that feature. Attribute them separately:
+  who wrote the feature, who wrote this fix.
+
+State it plainly and without hedging: "Not me — this is <contributor>'s PR. The
+feature it touches is pre-existing project code, yours." If Claude did write
+it, say so in the same breath, and in which session or PR.
+
+Never claim authorship you have not checked, and never let "we" blur it.
+
+## 3. Answer in this order
+
+The first thing on screen is a mental model the reader can hold without knowing
+the codebase. Not the authorship, not the file, not the constant. If they read
+only the first six lines, they should be able to state the bug back to you in
+their own words.
+
+1. **The model and the thesis.** Open with a concrete, everyday analogy for the
+   subsystem, then state the entire defect in ONE bolded sentence of plain
+   words. Then write "That's the whole thing. Everything below is just why."
+
+   **Hard budget for this section: zero `file:line`, zero constant names, zero
+   PR numbers, zero usernames, every sentence under 15 words.** If you cannot
+   write the thesis without an identifier in it, you do not yet understand the
+   change well enough to explain it. Go back to the diff.
+
+   Where the system has a small set of possible behaviours, enumerate the whole
+   set first, then say which one is wrong. Three rows of a table beats three
+   paragraphs:
+
+   > You ask: "how much of my quota is left?"
+   > There are only three honest answers:
+   >
+   > | Answer | Means |
+   > |---|---|
+   > | "You're at 95%" | Here's your data |
+   > | "You have no agents set up" | Nothing to report |
+   > | "I couldn't check just now" | Something went wrong |
+   >
+   > **The bug: it gives answer 2 when the truth is answer 3.**
+
+2. **Who wrote what.** Two or three lines. Still never skipped and never hedged
+   (§2), it just does not go first. A name and a PR number mean nothing until
+   the reader knows what the thing does.
+3. **The feature, and how someone actually uses it.** What is this capability
+   *for*? Give the real reason a person reaches for it ("your home drive is
+   small and you want the store on an external SSD"), not a restatement of its
+   name. One `file:line` anchor, at the end of the paragraph, not the start.
+4. **What you actually see.** Pure symptom, numbered, in the order the person
+   experiences it. **No mechanism in this section at all**, not a constant, not
+   a function name, not a reason. "The Plan-limits box disappears from your
+   sidebar" belongs here; "because the errors list was empty" does not.
+5. **Why it happens.** The mechanism, now that the symptom is concrete, with
+   `file:line` anchors so every claim can be checked.
+6. **What the change does about it** — including **the simpler fix it passed
+   over, and why**. There is almost always an obvious cheaper repair; naming it
+   and saying why the author went further is what turns "here is a diff" into
+   "here is a judgement I can agree or disagree with". For #331: you could tell
+   the one messy test to clean up after itself, but that leaves the next
+   contributor free to make the identical mistake. Also say what the change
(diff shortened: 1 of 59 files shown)


VasiHemanth:
Thanks for this. It's an unusually complete agent integration. The registration surface is fully covered (BUILDERS, `_KNOWN_AGENTS`, the quotas roster, billing mode, retention, the frontend `AgentKey` union), the parsing is defensive throughout, `~/.kimi/credentials/` is genuinely never read, and shipping 11 tests plus a summarizer adapter and a panel in one PR is more than I usually get. I appreciate the work and I want to land it.

One thing up front that shapes everything below: **I don't have access to Kimi Code myself.** I have a `~/.kimi` directory from an install, but no account and zero recorded sessions, so I can only check the config/registry side. I can't validate anything that depends on real `wire.jsonl` data. That means a couple of questions I can't answer without you.

## Two questions only your install can answer

**1. Are repeated `StatusUpdate` rows for the same `message_id` identical, or cumulative?**

`_scan_kimi_sessions` de-dupes **first-wins** (`main.py:4812`): once a `message_id` is in `seen_usage`, later samples are dropped. `test_scan_kimi_dedupes_status_updates_by_message_id` covers the case where the repeat is byte-identical, which is correct if Kimi re-sends a finished sample.

But if `StatusUpdate` is a streaming progress update, `output` grows with each emission for the same `message_id`, and first-wins would keep the smallest sample and undercount output tokens, silently.

Could you paste (redacted) the raw `StatusUpdate` rows for a single `message_id` from a real multi-turn `wire.jsonl`? Something like:

```
grep '"StatusUpdate"' wire.jsonl | head -20
```

If they are cumulative, switching to last-wins or a per-field `max()` fixes it. That change is a no-op under your stated assumption and a fix under the other, so it may be worth doing regardless.

**2. What does `default_model` actually resolve to on a working install?**

`_kimi_default_model()` falls back to `"kimi-for-coding"`, but that string isn't in `pricing.py`, so it prices at **$0.00 per token**. I checked:

```
calculate_cost("kimi-for-coding", 1_000_000, 1_000_000, 0, cache_creation_tokens=0)  ->  0.0
calculate_cost("kimi-k2", ...)                                                       ->  0.0
calculate_cost("kimi-k2.6", ...)                                                     ->  4.95
```

That matters because the fallback is reachable on a **default install**, not just a corrupt one. A fresh `~/.kimi/config.toml` here looks like:

```toml
default_model = ""
default_thinking = false

[models]

[providers]
```

`default_model` is an empty string and `[models]` is an empty table, so resolution walks straight to the fallback and every Kimi session is costed at zero with no error anywhere. Your test fixture writes `default_model = "kimi-k2.6"`, which is why the suite passes.

So: on a real logged-in install, is `default_model` populated, and does `[models."<id>"].model` map to something that exists in `pricing.py`? If Kimi's own model ids don't match our table, we may need an alias map rather than a fallback string.

## Changes you can make without me

These I verified directly and they don't need your install to fix.

**Blocking**

1. **`main.py:4720` can take down the whole `/sessions` scan.** The line is
   ```python
   entry = data.get("models", {}).get(model)
   ```
   with no `isinstance` check, unlike the guarded lookup eight lines above it. With a hand-edited `models = ["a", "b"]` in `config.toml` (valid TOML) this raises:
   ```
   AttributeError: 'list' object has no attribute 'get'
   ```
   It's raised from `main.py:4762`, which is **outside** the per-session `try`, and the call site `sessions.extend(_scan_kimi_sessions())` is unguarded like its siblings. So a malformed Kimi config 500s `/sessions` for all 19 other agents, not just Kimi. An `isinstance(models, dict)` guard closes it.

2. **Pick a fallback model that exists in `pricing.py`** (see question 2). Worth adding a fixture that writes the shipped default config verbatim and asserts `cost > 0`, so this specific case is pinned rather than only the happy path.

3. **Project attribution is `"unknown"` on a fresh install.** `_kimi_project_by_session()` requires `work_dirs[].last_session_id`, but that field is `null` in a `kimi.json` that hasn't completed a session, so the map comes back empty. Separately, the docstring at `main.py:4731` says the bucket dir "can't be reversed", which is true, but you don't need to reverse it. The bucket name appears to be plain `md5(work_dir_path)` with no trailing slash. You can check on your own install with:

   ```bash
   python3 -c "import hashlib,sys;print(hashlib.md5(sys.argv[1].encode()).hexdigest())" /your/work/dir
   ```

   and compare against the directory name under `~/.kimi/sessions/`. It matched exactly for the one `work_dir` in my `kimi.json`, so n=1 and I can't tell whether it's the raw path or a normalised/realpath form. If it holds, hashing known work dirs forward resolves **every** session for a project instead of only each directory's most recent one, which is a real improvement over the registry. Keeping the `last_session_id` lookup as a fallback seems right.

**Non-blocking**

4. **UPDATE.json contributor credit.** Per this repo's `CLAUDE.md`, significant external contributions get a credit line at the end of the relevant highlight's description: `Thanks to kthreatt for the contribution (PR #329).` Please add it; I'd rather it came from your PR than be added silently afterwards.

5. **UPDATE.json `href` points at a route the dashboard doesn't have.** `WhatsNewModal.tsx:106` renders any `href` starting with `/` as an in-app Next `<Link>`, and the frontend has no `/docs` route, so `/docs/supported-agents` 404s inside the app. `/agents/kimi` exists via `agents/[key]` and matches what the other agent entries use (`/agents/qoder`, `/agents/copilot`). (There's one pre-existing `/docs/...` href on `main` with the same problem, so this isn't yours alone, but let's not add a second.)

6. **Two doc gaps.** `KIMI_HOME` is a new TokenTelemetry-side override and isn't in the env-var table in `website/content/docs/reference/cli.mdx` (compare the `QODER_HOME` row). And `~/.kimi/sessions/` is missing from the scanned-paths list in `website/content/docs/quick-start.mdx`.

7. **PR description nit:** it says the model falls back to `kimi-k2`, but the code falls back to `kimi-for-coding`. Both price at 0.0, so it doesn't change the fix, but the description is stale.

8. **Minor, your call:** `_ARCHIVABLE` now includes `kimi`, which archives `wire.jsonl` only. The session title lives in the sibling `state.json`, so an archived Kimi session loses its title. Fine if that matches how `claude`/`codex` archives behave, just flagging it.

## Housekeeping

No CI has run on this branch, so nothing automated has looked at it yet. I couldn't run the backend suite in the environment I reviewed from (missing deps), so your "785 passed" is unverified on my side. The three findings above don't depend on the suite: I reproduced them against the config/pricing code directly.

None of this is a rejection. Items 1-3 are what stand between this and merge, and 1 and 2 are each a few lines. Happy to take it in whatever order suits you.

kthreatt:
Hello, 

I am working on your comments. Thank you for your response. But thanking to myself in my own PR is kinda weird. I cannot add it. Will fix rest of the issues.

VasiHemanth:
Maintainer edits were enabled, so I resolved the conflict with `main` on this branch rather than asking you to do it. Four commits: the merge, plus three follow-ups. The branch is mergeable again.

## What the review found correct

Worth saying explicitly, because two of these look like bugs and are not.

**The cache accounting is right, and the ZCode fix does not transfer.** ZCode (#362, merged yesterday) had cache reads billed twice, because its store reports `input` *inclusive* of `cache.read` while `calculate_cost` expects input already net of it. Kimi reports the opposite shape: `token_usage` is `{input_other, output, input_cache_read, input_cache_creation}`, and `input_other` is by name the non-cache portion. The scanner maps `input_other` to `tokens["input"]` and `input_cache_read` to `tokens["cached"]` separately, which is exactly what `calculate_cost` wants. Applying the ZCode fix here would have produced an undercount. Nothing to change.

**The work-dir hashing in `fix: harden Kimi session scanning` is what makes project attribution work at all.** `_kimi_project_by_session()` forward-hashes each `work_dirs` path to md5 and maps the bucket directory, falling back to `last_session_id`. I checked that against a real install: the single bucket under `~/.kimi/sessions/` matched `md5(work_dir)` exactly, and that `work_dirs` entry had **no** `last_session_id`. A mapping that trusted `last_session_id` alone would have left every session on that machine grouped as "unknown". The forward hash is carrying the whole feature.

**Malformed-input handling holds up.** Per-line `json.loads` in a `try`, a per-file `except Exception: continue`, `errors="replace"` on the read, and de-duplication by `payload.message_id` before summing, which matters because the same usage sample repeats across `StatusUpdate` rows. A truncated or half-written `wire.jsonl` degrades to a skipped session rather than a failed scan.

## The merge

26 conflicting hunks across 18 files, every one of them the same shape: `main` added ZCode to a roster and this branch added Kimi Code to the same roster. All resolved keep-both, with ZCode first so the diff reads as an append. No semantic change to either side: both agents keep the labels, hexes, icons, notes and captures their own branch gave them.

Four hunks needed more than concatenation because both sides had edited the same physical line: the `lucide-react` import in `agents.ts` (which now needs `SquareTerminal` for ZCode's meta entry *and* `Moon`), the `AgentKey` union, and the shared roster lines in `telemetry.py`, `test_quotas.py` and `test_harness_panels_agents.py`.

One thing worth flagging because it is the failure mode this kind of conflict hides. `isLightBrand` appears in both `AgentLogo.tsx` and `Badge.tsx`, and both copies conflicted and merged correctly. But the *comment* above `AgentLogo`'s copy did not conflict, so git kept this branch's version verbatim, and that version had never heard of ZCode. A fact spread across two adjacent lines gets half-updated, and nothing fails. Restored in the merge commit. Your comment wording was kept over main's in both files, incidentally, because "Monochrome brand marks (near-white ...; near-black Kimi)" covers both directions and main's "Near-white brand marks" is wrong for Kimi.

## Three fixes

**Kimi Code's real brand mark.** `@lobehub/icons` exports a `Kimi` mark and was already a dependency of both the app and the site, so the Lucide `Moon` placeholder was never needed. `Kimi` ships a full-colour variant, so it joins `COLOR_MARKS` next to Qoder. Its `COLOR_PRIMARY` is `#000`, the exact mirror of Qoder's `#fff`, which confirms the near-black hex and theme-aware `--agent-kimi` override you already had: the mark does need a light-mode flip, it just flips the other way from the near-white brands. `Moon` stays as the meta fallback in `agents.ts`.

**Counts and rosters.** The branch was written when the roster held 17 coding agents; ZCode made it 18 and Kimi makes it 19. The files that state a count, or spell the roster out in prose, are not files either branch edited, so they merged cleanly while going stale: the FAQ answer, the site metadata description, the docs index sentence, both `llms.txt` copies, and three landing-page counters. The convention is "N coding agents plus Hermes", so the coding count goes 18 to 19 and the total goes 19 to 20, which keeps the grid heading true against the 20 entries in `website/src/data/agents.ts`. Kimi Code also joins the billing-mode table, the harness-panel table and the troubleshooting path list, each of which had gained a ZCode row that Kimi never got.

**One capability claim the dashboard does not back.** `supported-agents.mdx` listed `subagents` for Kimi Code and said subagent runs are "folded into the parent session as delegation". The dashboard shows no such thing: `_scan_kimi_sessions` sets no `delegation` key, `kimi` is not in `_DELEGATION_CAPABLE_AGENTS`, and the session page does not request the delegation overlay for it, so `/sessions/{id}/delegation?agent=kimi` reports unsupported. The site's own `agents.ts` already omitted `subagents`, so the two surfaces contradicted each other. The text now claims only what is shown. Whether the wire log carries a spawn marker the scanner simply ignores is a separate question, and I have not asserted either way, since I had no recorded session to confirm the format against.

`UPDATE.json` gets a `2026-09-16` entry so the pre-push gate is satisfied. It is a new release rather than a sixth bullet on `2026-09-15`, which is already at the five the drawer renders, and the first highlight carries the contributor credit line.

## Verification

- Full backend suite: **893 passed**, with exactly the same 7 failures as `origin/main` on this machine (compared failure lists directly against a pristine `origin/main` worktree, not against a remembered baseline). No regressions; the branch adds 17 passing tests.
- `tsc --noEmit` clean in both `frontend/` and `website/`.
- `build_kimi()` against a real `~/.kimi` returns `installed: True` with the two sections the docs now describe: Configuration (max steps per turn, max retries per step) and Work directories. Credentials are reported as present and never opened, which the panel gets right.
- A segfault appeared once at interpreter teardown after the full run reported its summary. It did not reproduce in three consecutive runs of `test_kimi_scan.py`, and all tests had already passed, so I am recording it rather than attributing it to this branch.

Nothing needed from you. Thanks for the integration, and particularly for the hardening pass; the forward hash was the right call and the install I tested would not have worked without it.

PR #336, opened 2026-09-08
perf: serve production build instead of next dev (fixes ~1GB growth + slow trace loads)

slmingol:
Rebase of #187 onto current \`main\`.

## What

\`bin/cli.js\` started the frontend with \`next dev\` so everyone who clones and runs was using the Next.js dev server as their production runtime. This PR:

- Adds \`frontendBuildKey()\` (git HEAD SHA, falls back to package.json hash) to detect when a rebuild is needed
- Adds \`ensureFrontendBuild()\` which runs \`next build\` on first launch and whenever the key changes; subsequent launches skip it (~15ms start vs seconds)
- Switches from \`next dev\` to \`next start\` by default; adds \`--dev\` flag to opt back into HMR for repo contributors
- Keeps \`--no-open\` and \`--hostname host\` binding in dev mode for security

## Conflict resolution

The one conflicting file (\`bin/cli.js\`) had accumulated \`--no-open\`, \`noOpen\` flag, and Windows \`.cmd\` spawn handling since #187 was opened. Resolution kept all of HEAD's additions and merged \`--dev\` alongside \`--no-open\` in \`parseArgs\`; kept \`start(options)\` signature so \`main()\` dispatch is unchanged; added \`--hostname host\` back to the dev-mode args for localhost-only binding.

Closes #187

### DIFF
--- a/README.md
+++ b/README.md
@@ -314,7 +314,7 @@ This pattern is common when your agents (and their logs) run on a remote VPS or 
 NEXT_PUBLIC_API_BASE=http://localhost:8000 ./start.sh
 ```
 
-The `NEXT_PUBLIC_API_BASE` override tells the frontend to always talk to the backend at that address (instead of deriving it from the browser's window.location). It is inherited by the Next.js dev server.
+The `NEXT_PUBLIC_API_BASE` override tells the frontend to always talk to the backend at that address (instead of deriving it from the browser's window.location). It is baked into the frontend build, so setting, changing, or clearing it triggers a rebuild on the next `./start.sh`.
 
 **On your laptop:**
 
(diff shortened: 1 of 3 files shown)


slmingol:
Fixed in the latest commit: the duplicate `ensureFrontendBuild()` call (without `apiPort`) that appeared just before the frontend spawn was removed. The single correct call with `apiPort` at the top of the setup sequence remains.

slmingol:
Addressed the second Copilot review:

1. **`--host` inconsistency**: `next start` now receives `--hostname host` (same as `next dev`). Without it, `next start` was binding to all interfaces regardless of the operator's `--host` argument.

2. **Build key lockfile gap**: `frontendBuildKey()` fallback now hashes both `package.json` and `package-lock.json` (when present), so a lockfile-only change (e.g. `npm audit fix`) still triggers a rebuild in tarball installs without git metadata. Git-based installs are unaffected — git HEAD remains the primary key.

slmingol:
Fixed: moved `ensureFrontendBuild(apiPort)` to after `ensurePortsFree`. A port conflict now fails fast without triggering a 30-60s build.

PR #356, opened 2026-09-11
fix(quotas): report a cache-lock timeout as an error, not an empty result

### LINKED ISSUE #346: Quota cache lock timeout returns empty capabilities with no error, blanking the Plan-limits UI
Found by `/bug-audit` on 2026-09-10, auditing through `ea296438f0a3d89bf802da0d695c25936fbb8fd3`. `backend/quotas.py` is 1,441 lines, net-new since the last audit and not previously reviewed.

**FILE:** `backend/quotas.py:1314`
**SEVERITY:** high

## Finding

When the interprocess cache lock cannot be acquired, `collect()` returns a response with an empty `capabilities` map and an empty `errors` list — which every consumer reads as "nothing to show" rather than "we could not look".

`_CacheFileLock.__enter__` gives up after `CACHE_LOCK_TIMEOUT_SECONDS = 15`; the fallback path returns `self._wire(self.now(), [])` with `capabilities=None` → `{}` and `errors=[]`.

## Why the timeout is reachable in normal use

A worst-case forced refresh does roughly 11 outbound requests at a 10s timeout each (Codex 2, Gemini 2, Grok 3, plus four more providers). A menubar app and the dashboard backend racing on `~/.tokentelemetry/quotas.json` will exceed 15s. This is two of the app's own surfaces contending, not an exotic condition.

## What the user sees

The loser blocks for 15s and then returns a response indistinguishable from "no agents configured":

- `frontend/src/components/QuotaIndicator.tsx:49-51` computes `providers.length === 0 && unavailable.length === 0` and returns `null` — the whole Plan-limits control **disappears from the sidebar**
- `backend/menubar/presentation.py:157-178` finds `worst is None`; `_response_failure` returns `None` because `errors` is empty; renders `state: "no_data"`, title `◔ No quota data`

A user sitting at 95% of a weekly window sees no gauge and no error message at all.

## Reproduction

Held `_CacheFileLock` on a cache path from a separate process, then called `collect(force=True)` from the parent. It returned after 15.0s with:

```
providers={} capabilities={} errors=[]
```

## Suggested direction

A lock-acquisition failure is an error, not an empty result — put something in `errors` so the surfaces can distinguish "couldn't read" from "nothing configured". Serving the last cached snapshot rather than an empty response would also be strictly better than blanking the UI.

AmirF194:
`_CacheFileLock.__enter__` gives up after 15s and `collect()`'s fallback returns `self._wire(self.now(), [])`. The `errors` list stayed empty, so `_response_failure` (menubar) and the dashboard's `providers.length === 0` check can't tell "we couldn't look" from "nothing configured", exactly as the finding says.

Added an error entry to that fallback (`providerId: "quotaCache"`, same shape the provider-refresh path already uses) so a lock timeout is visible instead of silent. Left the "serve the last cached snapshot" half of the suggestion alone: `self._load(reload=True)` runs right before the fallback already, so a real previous snapshot does get served, this path just never said so when it couldn't refresh.

Added `test_service_reports_a_lock_timeout_as_an_error_not_an_empty_result`, forcing `_CacheFileLock.__enter__` to return `False` and checking the response. The new test fails without the change and passes with it. Full `backend/` suite: 827 passed vs 826 before, same 10 pre-existing failures either way (antigravity_cli, delegation, hook_push_cwd, update_check, none of them touch quotas.py).

Didn't touch the frontend; `QuotaIndicator.tsx` and `presentation.py` both already read `errors`/`providers` the way this response now populates them.

Fixes #346

### DIFF
--- a/backend/quotas.py
+++ b/backend/quotas.py
@@ -1317,7 +1317,10 @@ class QuotaService:
                     # means this read is either the prior complete cache or the
                     # next complete cache; never write a competing snapshot.
                     self._load(reload=True)
-                    return self._wire(self.now(), [])
+                    return self._wire(self.now(), [{
+                        "providerId": "quotaCache",
+                        "message": "Could not acquire the quota cache lock; showing the last cached snapshot.",
+                    }])
 
                 # Reload *inside* the process lock. A long-lived dashboard or
                 # menubar instance may have loaded a stale cache before another
--- a/backend/test_quotas.py
+++ b/backend/test_quotas.py
@@ -22,6 +22,7 @@ from quotas import (
     QuotaService,
     QuotaSnapshot,
     StaticQuotaProvider,
+    _CacheFileLock,
     default_quota_providers,
 )
 
@@ -302,6 +303,30 @@ def test_service_keeps_last_good_snapshot_when_a_refresh_fails(tmp_path):
     assert second["providers"]["codex"]["fetchedAt"] == first["providers"]["codex"]["fetchedAt"]
     assert second["errors"] == [{"providerId": "codex", "message": "Could not refresh quota data."}]
     assert second["capabilities"]["codex"]["state"] == "refreshFailed"
+
+
+def test_service_reports_a_lock_timeout_as_an_error_not_an_empty_result(tmp_path, monkeypatch):
+    class Provider:
+        provider_id = "codex"
+        display_name = "Codex"
+
+        def has_local_credentials(self):
+            return True
+
+        def refresh(self, now):
+            raise AssertionError("must not refresh while the cache lock is held elsewhere")
+
+    monkeypatch.setattr(_CacheFileLock, "__enter__", lambda self: False)
+    monkeypatch.setattr(_CacheFileLock, "__exit__", lambda self, *exc: None)
+
+    result = QuotaService([Provider()], cache_path=tmp_path / "quotas.json").collect(force=True)
+
+    assert result["providers"] == {}
+    assert result["capabilities"] == {}
+    assert result["errors"] == [{
+        "providerId": "quotaCache",
+        "message": "Could not acquire the quota cache lock; showing the last cached snapshot.",
+    }]
 
 
 def test_service_reports_not_signed_in_and_unsupported_harnesses(tmp_path):


VasiHemanth:
Thanks for this, and sorry to close it. Your diagnosis was right and it's carried forward, so let me explain exactly what happened rather than just closing silently.

You identified the real problem: a response with empty `capabilities` and an empty `errors` list is indistinguishable from "no agents configured", so every surface renders as though nothing is wrong. That framing is what made the next step obvious, and it survives in #357.

## Why it's being superseded rather than merged

Your PR reports the lock timeout. Looking at it with that framing in hand, the better question turned out to be why there is a 15-second timeout to report at all.

`_CacheFileLock` exists (703f397) so the dashboard and the menubar don't fetch the same quotas twice or overwrite a newer snapshot with an older one. A caller that loses the race does neither of those things: it reads the cache and returns. It was waiting the full `CACHE_LOCK_TIMEOUT_SECONDS` to learn that, then serving the same file it could have read immediately. The code even says so on the path that ignores it:

> "Another process is still collecting. Its atomic replace means this read is either the prior complete cache or the next complete cache; never write a competing snapshot."

So in #357 a background poll no longer waits at all. An explicit `POST /quotas/refresh` still does, because someone clicked a button and is watching. The lock's actual guarantees are untouched: the loser still never refreshes and never writes, and both guardrail tests from 703f397 pass unmodified.

Measured, those two paths take 30.12s against current `main` and 0.13s after.

## Your error entry is still there

It's kept, narrowed to the case that needs it: a held lock **and** no cached snapshot, which happens on a first run or a fresh data directory. That is the only time the response is genuinely indistinguishable from "no agents configured". When a cached snapshot does exist, it's now served immediately with no error, because nothing is wrong and every snapshot already carries `fetchedAt` / `expiresAt` / `stale` for the caller to judge freshness itself.

## One correction worth having

Your description says:

> "Didn't touch the frontend; `QuotaIndicator.tsx` and `presentation.py` both already read `errors`/`providers` the way this response now populates them."

Half of that holds. `presentation.py:294`'s `_response_failure` does read `errors[].message`, so the menubar half of your fix works as described.

`QuotaIndicator.tsx` does not. Line 51 gates on `providers.length === 0 && unavailable.length === 0`, and the file contains no reference to `errors` at all, so adding an error entry left the sidebar still rendering `null`. Worth knowing for next time; it's an easy one to get caught by, since `errors` is correctly typed in `lib/quotas.ts:30` and simply never consumed.

## Also

Your #355 is merged, and this is your second good catch in the same file in two days. The `from_wire` clock-skew fix was exactly right and is in `main` now. Please do keep sending these.

VasiHemanth:
Closing in favour of #357, which carries this forward. Reopen if that doesn't land.

AmirF194:
Good catch on QuotaIndicator.tsx, I read providers/errors as already wired there and they weren't. The no-wait poll in #357 is the better fix anyway, the timeout was never buying the loser's path anything.

Issue #39, opened 2026-06-02
[BUG] Unknown models


andresako:
## Describe the bug
 OpenCode sessions show the model as `unknown` even when the model is present in the local OpenCode SQLite database.
 
 On my machine, the current OpenCode parser still resolves **655 / 655 sessions** to `unknown`. The main issue is that it only looks for `message.data.model` on **assistant** messages, but in my local DB the model metadata is not stored there.
 
 ## To Reproduce
 1. Run `./start.sh`
 2. Use OpenCode so it writes session data to `~/.local/share/opencode/opencode.db`
 3. Open TokenTelemetry
 4. Navigate to the dashboard / analytics view, or open an OpenCode session
 5. See the model displayed as `unknown`
 
 ## Expected behavior
 TokenTelemetry should resolve the OpenCode model from the local SQLite data and display the real model name instead of `unknown`.
 
 At minimum, it should use the session-level model when available. Ideally, it should also expose multiple models when a single session used more than one.
 
 ## Environment
 - **OS**: macOS 26.5
 - **Node.js version**: v25.9.0
 - **Python version**: Python 3.12.3
 - **Agent(s) in use**: OpenCode
 - **TokenTelemetry version**: 1.0.0 (`f00dfb5`)
 
 ## Logs / Screenshots
 Relevant local DB observations:
 
 - Total OpenCode sessions checked: **655**
 - Sessions resolved as `unknown` by current logic: **655**
 - Assistant messages with model metadata: **0**
 - User messages with model metadata: **1899**
 - Sessions with a non-empty `session.model` value: **44**
 
 Example session rows:
 - `ses_195dcb072ffevuALMoUJFMAOlk | feat: create branch and apply changes (no commit)`
 - `ses_195dc46b8ffettjRME3k3CNghS | Find App Check implementation (@explore subagent)`
 
 Example of `session.model` content in the DB:
 ```json
 {"id":"claude-opus-4.6","providerID":"github-copilot","variant":"default"}
```

Example of a mixed-model session I found locally:

 - session: ses_195dcb072ffevuALMoUJFMAOlk
 - models seen: claude-opus-4.6, gpt-5.4

Additional context

The current parser appears to do this for OpenCode:

 - read message.data.model
 - only on role == "assistant"
 - fall back to modelID or providerID

That does not match the schema on my machine.

A more reliable approach would be:

 1. Use session.model as the primary source for the session-level model
[...]

PR #362, opened 2026-09-14
feat: ZCode (Z.ai) support — OpenCode-family SQLite store

Rub3nCT:
## Summary

Adds **ZCode** (Z.ai's coding CLI) as a fully supported agent. ZCode keeps its state in an OpenCode-family SQLite store at `~/.zcode/cli/db/db.sqlite`, so this PR mirrors the existing OpenCode integration: discovery helpers, a session scanner, trace + delegation endpoints, GLM-5.3 pricing, frontend registration, and the website/roster surfaces — 13 commits, each thematic.

> **Maintainer note (added before merge).** Six follow-up commits were pushed onto this branch via maintainer edits, so the merged PR is 20 commits rather than 13. They add the harness panel and the brand mark this PR deferred, close the docs gaps, and fix two defects found while reviewing. Sections below marked **Updated** describe what actually merged; the contributor's original text is otherwise unchanged. Details in the [review comment](#issuecomment-5683461254).

## Type of change
- [x] New feature (new agent support, new dashboard view, etc.)

## What it reads (schema verified against a live install)

- `session(id, directory, title, parent_id, time_created, time_updated)` — projects, titles, subagent links
- `message.data` (JSON) — role, modelID, providerID (`builtin:zai-*`), mode, per-message tokens; non-assistant rows carry `model` as a `{providerID, modelID}` dict, resolved like OpenCode's #39
- `part.data` (JSON) — text, tool calls, and `step-finish` token blocks with the **same field shapes as OpenCode**, so tokens are summed with identical semantics (input summed as-is — inclusive of cache reads; `cached` = high-water mark; cache writes cumulative) and the two same-schema agents stay comparable
- `todo` — plan artifacts

**Updated:** the `step-finish` input is indeed inclusive of `cache.read`, but `calculate_cost` expects input *net* of cache reads, so passing the gross figure alongside `cached` billed those tokens twice. The scanner now nets each step before costing. The OpenCode scanner has the same shape and was deliberately left alone here so the two same-schema agents don't diverge inside this PR; it needs its own change.

The panel additionally reads `model_usage`, `turn_usage`, `tool_usage` and `local_setting`, which the session scan never touches.

Children (`session.parent_id`) are annotated on the parent as delegation, never re-summed (count-once invariant). The store is opened read-only (`mode=ro`, WAL-safe, 1 s timeout); a corrupt or locked store degrades to a debug log, never an empty agent. `ZCODE_DATA_DIR` relocates the store like `OPENCODE_DATA_DIR` does.

**Updated:** `_zcode_db_path()` and `_zcode_dbs()` called `Path.exists()` without the `OSError` guard the OpenCode helpers carry. Since `ZCODE_DB` resolves at module import, a stale mount or an unreadable `~/.zcode` made `import main` raise and stopped the backend booting for *every* agent. On Python 3.9 to 3.12, `pathlib`'s `exists()` re-raises EACCES and ESTALE; only 3.13+ swallows them. Now mirrors the OpenCode guards.

## Cost

Sessions billed through the coding plan report `cost: 0` from ZCode itself, so TokenTelemetry estimates API-equivalent cost at Z.ai's list rates: `glm-5.3` = 1.4/4.4/0.26 and `glm-5.3-flash` = 0.15/0.5/0.03 USD per Mtok (Z.ai's official prices, corroborated by LiteLLM and the major aggregators; models.dev's `zai` section currently lists a half-price promo tier). Note the estimate inherits the OpenCode scanner's cache semantics by design — if you ever revisit inclusive-input billing, both agents should move together.

**Updated:** that revisit happened here for ZCode. Measured against a real store, the fleet estimate fell from **$4.4595 to $0.6671** once cache reads stopped being counted twice. The lower figure is the correct one. The `cached` high-water-mark still understates against summing per-step cache reads; that is left as-is for OpenCode parity, so the corrected number is a floor.

## What is deliberately not here

- ~~**No harness panel yet**~~ — **now included.** The claim that ZCode's store carries nothing the session scan doesn't already show did not hold: `model_usage`, `turn_usage`, `tool_usage` and `local_setting` give five sections the generic session view cannot, including per-request reasoning level (no other supported agent records it), turn latency and context overflows, and ZCode's own read-only/destructive tool labels with the approval each call received. `zcode` moves from `harness_panels.PLANNED` into `BUILDERS`, and `PLANNED` is now empty. Token totals are deliberately excluded so the panel cannot drift from the session scan.
- **No plan-quota gauge** — coding-plan usage lives in the Z.ai account API; nothing local reports it. Static quota-provider entry, same as Qoder. **Still true.** `GET https://api.z.ai/api/monitor/usage/quota/limit` looks well established and returns a 5-hour rolling and a weekly window with `nextResetTime`, which maps onto the existing gauges, but every implementation found authenticates with a *coding-plan* key and the test machine runs `builtin:zai-start-plan`. Rather than ship a gauge that guesses, the static entry stays until the tier caps and the accepted key type are confirmed.
- ~~**UPDATE.json**~~ — **now included**, tagged `2026-09-15` with five highlights. The cost correction gets its own entry, since a ZCode estimate that drops 85% overnight reads as a bug unless the release notes say the old number was the wrong one.

## How to test

1. Run `./start.sh` on a machine with ZCode history
2. The Agents row shows a **ZCode** tile with Z.ai's own mark; sessions appear under their real projects with `GLM-5.3-Flash` / `GLM-5.3` and non-default costs
3. Open a session — waterfall trace (user/assistant/reasoning/tool events, per-step tokens); delegated sessions show linked children, each with its own generated subagent mark
4. `/agents/zcode/panel` returns five sections; the quota page shows ZCode's static entry

Verified live against a real store (127 sessions, ~798M input tokens): scanner totals within 0.5% of ZCode's own `model_usage` ledger, zero model-less sessions, delegation linkage correct on 54 child sessions.

## Checklist
- [x] I tested this locally with `./start.sh` (backend API + scans verified against a live store; frontend and website builds pass)
- [x] PR is focused on one change (ZCode support; the two analytics/hermes hermeticity fixes it forced are part of the same story)
- [x] README, CHANGELOG and website updated

## Test evidence

Full backend suite: **845 passed**, with exactly the same 16 failures as `main` on this Windows/Python 3.14 machine (verified failure-list parity against a pristine `main` worktree — all 16 are environment-sensitive tests unrelated to this change, e.g. `test_hook_push_cwd` and the FastAPI `Query`-default leak in `test_analytics_ecosystem_aggregates`). New coverage: 20 tests across discovery, scanner, traces, delegation and pricing; every hermetic isolation helper that pins `OPENCODE_DB` now pins `ZCODE_DB` too, so full-scan tests stay hermetic on machines with a real ZCode store.

**Updated:** the follow-ups add `test_zcode_panel.py` (17 tests) plus a regression test per fix. Across the ZCode and panel suites, 187 pass on macOS / Python 3.14 with four failures verified pre-existing against a pristine pre-merge worktree (three Claude/Grok panel tests that read the real machine, plus the same `Query`-default leak named above). `tsc --noEmit` is clean in both `frontend/` and `website/`. The panel isolation gap this surfaced is covered: `hp_paths.zcode_dir` is now pinned in the every-agent isolation test, so a machine with a real `~/.zcode` cannot build a live panel inside it.

## Related issue
Closes #N/A (no tracking issue found for ZCode)


VasiHemanth:
Thanks for this. Maintainer edits were enabled, so I pushed five follow-up commits onto this branch rather than opening a competing PR. Summary of what changed, since some of it touches code you wrote:

**ZCode's real brand mark.** `@lobehub/icons` ships a `ZAI` component and was already a dependency of both the app and the site, so the Lucide `SquareTerminal` fallback and the placeholder rose tint are gone. It takes Qoder's neutral treatment instead: a near-white hex plus a theme-aware `--agent-zcode` that flips in light mode. Both copies of the `isLightBrand` predicate were updated, since `AgentLogo` and `Badge` each keep their own.

**A harness panel.** The registry note said ZCode's store carries nothing the session scan doesn't already show. That turned out not to hold: alongside the OpenCode-family `session`/`message`/`part` tables, ZCode keeps `model_usage`, `turn_usage`, `tool_usage` and `local_setting`. Those give five sections the generic session view can't: per-request reasoning level (no other supported agent records it), turn latency and context overflows, and ZCode's own read-only/destructive classification with the approval each call received. `zcode` moves from `PLANNED` into `BUILDERS`. Token totals are deliberately left out so the panel can't drift from the session scan.

**Docs consistency.** The FAQ still said "Seventeen coding agents" and omitted ZCode while the site metadata already said 18. That plus the docs index, quick-start, billing, dashboard, troubleshooting, CLI reference and both `llms.txt` files.

**Two fixes.** Both are worth calling out:

1. `_zcode_db_path()` and `_zcode_dbs()` called `Path.exists()` without the `OSError` guard the OpenCode helpers deliberately carry. Because `ZCODE_DB` is resolved at module import, a stale mount or an unreadable `~/.zcode` made `import main` raise, so the backend refused to start for *every* agent, not just ZCode. On Python 3.9 to 3.12, `pathlib`'s `exists()` re-raises EACCES and ESTALE; only 3.13+ swallows them. Now mirrors the OpenCode guards exactly.

2. Cache reads were billed twice. `calculate_cost` expects input *net* of cache reads and adds `cached` on top, and the scanner passed the gross `input` from `step-finish` alongside `cached`. The docstring already said that input is inclusive of `cache.read`, so the two disagreed. Measured against a real store: fleet cost went from $4.4595 to $0.6671. **The OpenCode scanner has the same shape**, and I left it untouched so the two same-schema agents don't diverge inside this PR. It needs its own change.

**Subagent sigils** on the delegated-work rows, so linked child sessions are distinguishable at a glance instead of a column of near-identical ids.

`UPDATE.json` now has an entry, so the pre-push gate is satisfied.

---

One thing I could not settle, and you may be able to.

A live plan-limits gauge for ZCode looks feasible. `GET https://api.z.ai/api/monitor/usage/quota/limit` appears well established across several independent implementations, returning `data.limits[]` with a 5-hour rolling window (`unit=3, number=5`) and a weekly one (`unit=6, number=1`), each carrying `usage` as the cap, `currentValue` as consumed, and `nextResetTime`. That maps directly onto the session and weekly gauges the dashboard already draws for other agents.

Two questions if you have a ZCode subscription:

1. **Are the 5-hour and weekly caps published for your plan tier?** If they are fixed per tier, we can pair them with usage derived from the local store (`model_usage.started_at` is enough) and draw the gauge with no network call at all. That would be the preferred outcome here, since the project avoids adding outbound calls from user machines where it can.

2. **Which key type does that endpoint accept?** Every implementation we found authenticates with a coding-plan key. The machine this was tested on runs `builtin:zai-start-plan`, and the one start-plan project we found used `zcode.z.ai/api/v1/zcode-plan/billing/balance` instead, which returns unit balances rather than percentage windows. Whether a start-plan token works against the monitor endpoint is untested.

Please don't share a key or any account details. Tier names and published limits are all that's needed.

The quota entry stays static until that's answered, rather than shipping a gauge that guesses at the numbers.

Issue #353, opened 2026-09-10
/dsh/lifecycle computes its summary over the capped tail, under-reporting failed transitions with no true total


VasiHemanth:
Found by `/bug-audit` on 2026-09-10, auditing through `ea296438f0a3d89bf802da0d695c25936fbb8fd3`.

**FILE:** `backend/main.py:5337` (the slice) and `backend/main.py:5395-5400` (summary computed on the sliced list)
**SEVERITY:** high

## Finding

`/dsh/lifecycle` slices to the last `limit` (default 500) transitions **before** computing the summary, so `failed` / `reloads` / `unloads` / `transitions` and the per-plugin rollup are all derived from the truncated tail — and no field anywhere in the response reports the true, unclipped total.

```python
# _dsh_lifecycle_events, main.py:5337
return out[-limit:] if limit and len(out) > limit else out

# endpoint, main.py:5395-5400
events = _dsh_lifecycle_events(since_ms=since, until_ms=until, limit=limit)
return {
    "installed": DSH_LIFECYCLE_FILE.exists(),
    "correlation": correlation,
    "events": events,
    **_dsh_lifecycle_summary(events),      # ← summary over the ALREADY-TRUNCATED list
}
```

This is the same shape as the bug that motivated this audit (PR #131): a cap applied before an aggregate is computed, with the capped figure presented as complete.

## Scenario

A DSH plugin has 550 failed-load transitions early in its history, followed by 50 healthy ones after a fix — 600 records in `dsh_lifecycle.jsonl`.

`GET /dsh/lifecycle` with the default `limit=500` returns `transitions: 500, failed: 450` instead of the true `transitions: 600, failed: 550`. A 100-count silent undercount, with no signal that the window was clipped.

Note the direction of the error: because the slice keeps the *tail*, it discards the oldest records — so a plugin whose failures are historical will under-report exactly the failures a user is looking for.

## Reproduction

Ran `_dsh_lifecycle_events` / `_dsh_lifecycle_summary` (bodies copied verbatim from `backend/main.py`) against a synthetic 600-line lifecycle log:

```
true summary                      : {'transitions': 600, 'failed': 550}
same functions at default limit=500: {'transitions': 500, 'failed': 450}
```

## Suggested direction

[...]

PR #368, opened 2026-09-17
fix(scan): require exact mtime match before trusting a cached parse

AmirF194:
`write_cache` always stores `_mtime = source_mtime`, so `read_cache`'s `cached_mtime >= source_mtime` check only diverges from equality when a later reading of `source_mtime` comes back smaller than what was recorded, which means the clock went backwards (NTP step, NAS/SMB skew) since the write. `>=` treats that as still fresh, so the payload from that one parse is served forever even as real appends keep landing on disk. Switched the comparison to `==`, the only relationship that actually means "unchanged since write."

One thing this doesn't fix: on a coarse-resolution filesystem (exFAT, many SMB mounts), an append landing inside the same tick as the parse keeps the same mtime, so it's still a hit with `==` too. That's a separate problem, mtime alone can't tell those two states apart, and it would need a stronger signal (size, line count) to close.

Added `test_scan_cache.py`, three cases: unchanged mtime still hits, an advancing mtime misses (both pass on main), and a backwards-moving mtime misses (fails on main, it still returns the stale payload there). Ran the full backend suite on python 3.11, 902 passed / 10 failed, and the same 10 fail on main too (antigravity_cli, delegation, hook_push_cwd, update_check), unrelated to this file. No CI job runs pytest on this repo, so this is the only check that ran.


AmirF194:
Same fix as #374 in scan_cache.py (== instead of >= on the mtime check), opened three days after this one. #374 also folds in #353's lifecycle-summary bug. Your call which one lands, didn't want the overlap to go unmentioned.

Issue #351, opened 2026-09-10
scan_cache accepts cached_mtime >= source_mtime, pinning a stale parse and freezing a session forever


VasiHemanth:
Found by `/bug-audit` on 2026-09-10, auditing through `ea296438f0a3d89bf802da0d695c25936fbb8fd3`.

**FILE:** `backend/scan_cache.py:117`
**SEVERITY:** high

## Finding

`read_cache` accepts any payload whose stored `_mtime` is **greater-than-or-equal** to the source mtime, so once a source file's mtime is non-monotonic the cached parse is pinned and every later append is invisible to both the live scan and the durable store.

`write_cache` stores `_mtime = source_mtime`; `read_cache` returns the payload when `cached_mtime >= source_mtime`.

For Claude, `source_mtime` is `max()` over the parent transcript plus every subagent and workflow file (`backend/main.py:6638-6659`), so it is only as trustworthy as the least monotonic clock touching any of them.

## Trigger, without user error

A home directory on an SMB/NFS share or a NAS whose server clock runs ahead of the client, or an NTP step-back after the machine ran fast.

A session parsed while the mtime source was ahead stores a future `_mtime`. Every subsequent append produces a smaller mtime, so the cache hits forever and the session freezes at the tokens it had at that moment. Because a cache hit sets `stub=False` (`backend/main.py:6666`, `7074`), the frozen undercount is written through the **non-stub** conflict clause into `history.db` on every scan.

## The equality case needs no clock anomaly at all

`>=` also accepts an *unchanged* mtime. On a coarse-resolution filesystem (exFAT/FAT32 at 2s on an external drive, many SMB mounts at 1s), any turns appended inside the same tick as the parse are permanently lost if the session then goes idle.

The module docstring claims the in-payload `_mtime` makes this "deterministic under coarse filesystem mtime resolution", which is backwards — coarse resolution is exactly what `>=` cannot survive.

## Reproduction

Sandboxed `CLAUDE_DIR`, real `_scan_sessions_sync` + `upsert_sessions`. Set the transcript's mtime to `now + 86400` (machine one day fast), scanned:

```
Scan 1: LIVE total=3000 | DB total=3000
append a second assistant turn worth 140,000 tokens, mtime back to corrected wall clock
Scan 2: file_lines=3  LIVE total=3000 | DB total=3000
Scan 3: file_lines=3  LIVE total=3000 | DB total=3000
```

[...]

PR #371, opened 2026-09-20
fix(dsh): compute lifecycle summary over the full log, not the display page

### LINKED ISSUE #353: /dsh/lifecycle computes its summary over the capped tail, under-reporting failed transitions with no true total

AmirF194:
`_dsh_lifecycle_events` reads the whole plugin lifecycle log, sorts it, then keeps only the newest `limit` (default 500) entries before `_dsh_lifecycle_summary` ever sees them. So `transitions`/`failed`/`reloads`/`unloads` are counted over that trimmed tail, not the real log: a plugin whose failures cluster early (say 550 failed loads, then 50 healthy ones after a fix) reports 500 transitions and 450 failed instead of the true 600 and 550, with nothing in the response saying the window was clipped.

Fix keeps the cap where it belongs, on the `events` list returned for display, but computes the summary over the full unclipped list first. `_dsh_lifecycle_events` no longer truncates at all; the endpoint slices the last `limit` events for the `events` field after building the summary from the complete set.

Added a test that reproduces the 550/50 split and checks both `_dsh_lifecycle_summary` directly and the `/dsh/lifecycle` endpoint. It reports 500/450 without the change and 600/550 with it. Also added a control test under the cap (2 events) to confirm the common case is unchanged. Ran the full `backend/` suite in a `python:3.11-slim` container: 911 passed vs 908 before, same 9 pre-existing failures either way (antigravity_cli symlinks, hook_push_cwd git tests, update_check preferences), none of them touch this file. No test/lint CI job exists on this repo, so this suite is the only check.

Fixes #353


AmirF194:
#374 fixes this too, bundled into the same PR as the scan_cache mtime fix on #351/#368. Different shape (keeps a limit=None call here vs this PR dropping the limit param entirely) but the same effect. Your call which one lands.

PR #373, opened 2026-09-20
fix(codex): reset _read_ok on a mid-parse failure, not just a failed open

### LINKED ISSUE #350: Codex partial rollout parse is marked stub=False and blind-replaces the durable row (98.75% loss reproduced)
Found by `/bug-audit` on 2026-09-10, auditing through `ea296438f0a3d89bf802da0d695c25936fbb8fd3`.

**FILE:** `backend/main.py:7264`
**SEVERITY:** high

## Finding

A read/parse error on one of a Codex session's rollout files is swallowed per-file, yet the truncated result is still marked `stub=False`, written to the sidecar cache as authoritative, and blind-replaced into the durable history row.

`for rollout_file in rollout_files:` (`backend/main.py:7111`) wraps the entire per-file body in `try: … except Exception: pass`. Control then falls straight through to `backend/main.py:7298-7300`, which does `scan_cache.write_cache(...)` and `sess["stub"] = False`.

Triggers that need no user error:

- `OSError` — EACCES, a Windows lock, a file removed mid-scan
- `AttributeError`/`KeyError` from schema drift — `data["payload"].get(...)` at `backend/main.py:7119` sits **outside** the per-line `json.loads` guard, so a `payload` that is a string rather than a dict raises

## Why the loss is large

Codex maps several rollout files to one session id (`backend/main.py:7027-7033`, uuid recovered from the filename) and accumulates tokens with `max()` over cumulative snapshots (`backend/main.py:7203-7206`). Losing the **newest** file reverts the session to the older file's smaller totals.

Because `stub` is `False`, `upsert_sessions` takes the non-stub conflict clause (`backend/history_store.py:235-240`) — `input=excluded.input, output=excluded.output, total=excluded.total, cost=excluded.cost`, a straight replace with no regression guard. The durable rollup, whose entire purpose is to outlive source loss, is overwritten with the truncated numbers.

## Reproduction

Sandboxed `CODEX_DIR` with two rollout files for one session id (older 15,000 tokens, newer 1,200,000):

```
Scan 1: LIVE total=1200000 cost=4.0125 stub=False | DB total=1200000 DB cost=4.0125
chmod 000 on the newer rollout, bump mtime to force a cache miss
Scan 2: LIVE total=15000   cost=0.0603 stub=False | DB total=15000   DB cost=0.0603
```

A **98.75% loss** written straight through the store, with the truncated parse cached in the sidecar so it persists until the source mtime moves again.

## The root gap

A scanner currently has no way to say *"this parse was incomplete"*. `stub` is the only signal, and a mid-file abort never sets it.

Note that `MAX()` in the conflict clause is **not** the fix — CACHE_VERSION 11's dedupe fix legitimately lowered totals, so a monotonic guard would have blocked a correct repair. The scanner has to mark the parse partial.

AmirF194:
Did what your comment on #350 suggested: extended `_read_ok` into a real parse-was-complete signal instead of adding a second flag. It resets to `False` inside the except block that already logs the read failure, so a mid-parse failure on a later file blocks the durable write for that pass instead of letting an earlier file's success carry it through.

Added a test with two rollout files: an older complete one, and a newer one that opens fine then hits a `session_meta` record with no `payload`. On main it writes stub=False using just the older file's 15000 tokens; with the fix it stays a stub. Full backend suite is otherwise unaffected, same 9 pre-existing failures either way (antigravity_cli symlinks, hook_push_cwd git worktree tests, update_check preferences), none touching this file.

No test/lint CI runs on this repo, so this suite is what I've got.

Fixes #350


AmirF194:
#375 fixes this same bug with a separate _partial_parse flag instead of resetting _read_ok directly, opened about two and a half hours after this one. It also bundles an unrelated quota sign-out fix. Your call which shape you'd rather keep.

